Issue #41, opened 2022-07-30
HttpError when Switch repo to same repository


gycgabriel:
**Describe the bug**
<img width="296" alt="image" src="https://user-images.githubusercontent.com/77230723/181905149-98479c60-87f9-428b-a33f-b2a6d7bc95a6.png">

**To Reproduce**

1. Click submit on switch repo without changing url
2. Click submit repeatedly before issues are fully loaded

**Expected behavior**
Check if repository is same, if it is do not refetch.

cheehongw:
The snackbar was caused by an uncaught error bubbling all the way up to the app level. 

![image](https://user-images.githubusercontent.com/72195240/216824312-735f78db-ea37-445f-b958-836d8acef634.png)

Investigations of the error message indicate that this was caused by the request made in `GithubService::getIssuesAPICall`.
An error in `Promise apiCall` is likely the source of this error
```ts
 private getIssuesAPICall(filter: RestGithubIssueFilter, pageNumber: number): Observable<GithubResponse<GithubIssue[]>> {
    const apiCall: Promise<GithubResponse<GithubIssue[]>> = octokit.issues.listForRepo({
      ...filter,
      owner: ORG_NAME,
      repo: REPO,
      sort: 'created',
      direction: 'desc',
      per_page: 100,
      page: pageNumber,
      headers: { 'If-None-Match': this.issuesCacheManager.getEtagFor(pageNumber) },
      state: 'all'
    });
    const apiCall$ = from(apiCall);   // <--- this line?
    return apiCall$.pipe(
      catchError((err) => {
        return of(this.issuesCacheManager.get(pageNumber));
      })
    );
  }
  ```

cheehongw:
We are running on rxjs 6.5.3. `catchError` does not appear to catch an error on an observable created from a promise.

Same problem has been observed elsewhere for rxjs 6.5.1:
https://stackoverflow.com/questions/59032093/rxjs-catcherror-operator-does-not-catch-an-error-on-an-observable-created-from-a

PR #60, opened 2023-01-26
Provide better messages for invalid repositories

### LINKED ISSUE #47: Error messages for invalid repo not friendly
When attempting to switch between different repos, an incorrect repo name can cause errors. Some of these errors are not handled well. Some of the error messages include:

- Invalid repo name (eg. CATcher-org/hello):
   - Uncaught (in promise): HttpError: Not Found HttpError: Not Found at https://catcher-org.github.io/WATcher/main-es2015.1461b2c7aaa1136b20ba.js:2:427803

- Invalid repo name without `/` (eg. CATcher-org)
  - Cannot read properties of undefined (reading 'owner')
 
We should aim to display more friendly error messages like "Invalid repo name provided."

vigneshsankariyer1234567890:
### Summary:
Fixes #47. Originally, if a repo name with the wrong name or organisation was provided, it was difficult and not user friendly to see the what caused the error. Furthermore, certain methods which requested issues or milestones from Github did not always catch the subsequent errors that were thrown. 

### Changes Made:
* Added better messages to errors thrown

### Commit Message:

```
Let's provide better messages for invalid repositories
```


chunweii (commented):
A couple of error messages are still not properly handled.
Some examples:
- `hello`
- (blank)

Furthermore, many error messages will pop up at once, and the error message that would be shown is not deterministic. Sometimes, the old verbose error message still pops up

vigneshsankariyer1234567890:
Many of the HTTP errors arise from the fact that the repository is not valid in the first place. Instead of sending multiple requests at once, what if we first check if the repository exists, and if it doesn't, throw an error immediately? There is already a method that does that under `github.service`

chunweii:
> Many of the HTTP errors arise from the fact that the repository is not valid in the first place.
> 
> Instead of sending multiple requests at once, what if we first check if the repository exists, and if it doesn't, throw an error immediately? There is already a method that does that under `github.service`. This would change it's behaviour from receiving and displaying errors in a non-deterministic manner, to just displaying a single error. @chunweii

Yes, I think that is a good idea!

damithc:
Also try not to increase the number of API calls, if it can be avoided. Even a slight increase per student can result in a big increase of API calls during critical times such as start of PE. Might result in unforeseen problems due to GitHub anti flooding measures.

src/app/auth/confirm-login/confirm-login.component.ts
```diff
@@ -58,19 +55,16 @@ export class ConfirmLoginComponent implements OnInit {
...
-        () => {
-          this.handleAuthSuccess();
-        },
-        (error) => {
-          this.authService.changeAuthState(AuthState.NotAuthenticated);
-          this.errorHandlingService.handleError(error);
-          this.logger.info(`ConfirmLoginComponent: Completion of login process failed with an error: ${error}`);
-        }
-      );
+    this.userService.createUserModel(this.username).subscribe(
```
chunweii, line 58: Why is the `.pipe(flatMap(() => this.githubEventService.setLatestChangeEvent()))` line removed here?

vigneshsankariyer1234567890: Apologies, I was experimenting with the method to understand what it produced and I must have forgot to reinsert it. I'm still not too sure what it does but I will reinstate the call.

chunweii:
I am still getting long and unfriendly error messages like

image

This issue is more difficult than it seems. It could be linked to #41 and #65. It requires more investigation

vigneshsankariyer1234567890:
I believe the errors have come about due to the new changes, refactoring of the repo change form to Header, and some of the errors which are thrown by PhaseService are not caught.

> I am still getting long and unfriendly error messages like
> 
> image
> 
> This issue is more difficult than it seems. It could be linked to #41 and #65. It requires more investigation

Since errors that are thrown by `PhaseService.changeCurrentRepository()` are not caught, we continue to get these errors. I have added changes that tackle this issue.

src/app/core/services/phase.service.ts
```diff
@@ -76,9 +76,22 @@ export class PhaseService {
    * If on Issue Dashboard, add previously visited repositories to otherRepos.
    * @param repo New current repository
    */
-  changeCurrentRepository(repo: Repo): void {
+  async changeCurrentRepository(repo: Repo) {
+    if (!repo) {
+      throw new Error('Invalid repo name. Please provide repo name in the format Org/Repo.');
+    }
```
cheehongw, line 82: maybe this error for a invalid repo string should be thrown in `Repo::of`?

see line 19 of `repo.model.ts`

src/app/shared/layout/header.component.ts
```diff
@@ -37,7 +36,7 @@ export class HeaderComponent implements OnInit {
   private readonly noButtonDialogMessage = "No, I don't wish to log out";
 
   /** Model for the displayed repository name */
-  currentRepo: string = '';
+  currentRepo = '';
```
cheehongw, line 39: why remove?

src/app/auth/confirm-login/confirm-login.component.ts
```diff
@@ -68,7 +68,7 @@ export class ConfirmLoginComponent implements OnInit {
         (error) => {
           this.authService.changeAuthState(AuthState.NotAuthenticated);
           this.errorHandlingService.handleError(error);
-          this.logger.info(`ConfirmLoginComponent: Completion of login process failed with an error: ${error}`);
+          this.logger.info(`Completion of login process failed with an error: ${error}`);
```
cheehongw, line 71: follow the `codefile: action` format

src/app/core/services/phase.service.ts
```diff
@@ -76,9 +76,22 @@ export class PhaseService {
...
 
+    const isValidRepository = await this.githubService
+      .isRepositoryPresent(repo.owner, repo.name)
+      .toPromise()
+      .then((isValidRepository) => isValidRepository);
+
+    if (!isValidRepository) {
+      throw new Error('Invalid repo. Please check your organisation and repo name.');
+    }
+
```
cheehongw, line 94: invoke `ErrorHandlingService` here to raise the error snackbar instead of throwing an `Error` to the calling function?

src/app/issues-viewer/issues-viewer.component.ts
```diff
@@ -101,7 +96,7 @@ export class IssuesViewerComponent implements OnInit, AfterViewInit, OnDestroy {
     this.githubService.getUsersAssignable().subscribe((x) => (this.assignees = x));
 
     // Fetch issues
-    this.issueService.reloadAllIssues();
+    this.issueService.reloadAllIssues().pipe(catchError((err) => 'Failed to fetch issues for repo. Check your repo name.'));
```
cheehongw, line 99: I don't think this line would do anything.
Maybe you intended to call the logger in the callback function?

src/app/issues-viewer/label-chip-bar/label-chip-bar.component.ts
```diff
@@ -42,7 +42,9 @@ export class LabelChipBarComponent implements OnInit {
       (response) => {
         this.logger.debug('LabelChipBarComponent: Fetched labels from Github');
       },
-      (err) => {},
+      (err) => {
+        throw new Error('Failed to fetch labels from Github.');
+      },
```
cheehongw, line 47: will this error be caught?
maybe it's better to invoke ErrorHandlingService here instead of throwing an error to the calling function above

src/app/core/services/phase.service.ts
vigneshsankariyer1234567890: if this is thrown in `Repo::of`, I think it's highly likely that there will be a need for wide-scale refactoring through out the app (which I think deserves an issue for itself). Let's do work on this as a separate issue

src/app/shared/layout/header.component.ts
vigneshsankariyer1234567890: By current linting rules, the type of currentRepo can be inferred as a string just by its declaration as ''. In this case, additional type annotations do not make any difference.

src/app/auth/confirm-login/confirm-login.component.ts
vigneshsankariyer1234567890: good catch, will change accordingly

src/app/core/services/phase.service.ts
vigneshsankariyer1234567890: I think this increases coupling between 2 unrelated parts of the programme (`PhaseService` is a "backend" service, whereas `ErrorHandlingService` is a "frontend" service). IMO it makes more sense to throw here and catch in the component instead, since `PhaseService` should not have the responsibility of initiating the display of the Error.

src/app/issues-viewer/issues-viewer.component.ts
vigneshsankariyer1234567890: thank you for catching this, this was to be deleted :)

src/app/issues-viewer/label-chip-bar/label-chip-bar.component.ts
vigneshsankariyer1234567890: i agree, will change to invoking ErrorHandlingService here

vigneshsankariyer1234567890:
PR has become super messy, closing it and reopening with another PR instead

PR #61, opened 2023-01-28
Modify card view to move milestone flag position

### LINKED ISSUE #44: Show milestone and labels in the same line
image

Suggestion: they can be shown in the same line, to save vertical space.

Echomo-Xinyu:
### Summary:

Fixes #44 

### Changes Made:

Modify card view to move milestone flag to the same line as GitHub labels

### Commit Message:

```
Let's modify card view to move milestone flag position for better view.
```


kkangs0226 (approved):
Good work! 👍

However, we might need to think of another design in the future, because due to the size of the cards, the labels will still overflow into the next line. 
image
image

We can dicuss this further.

PR #70, opened 2023-02-05
Fix uncaught httpError arising from promise

### LINKED ISSUE #41: HttpError when Switch repo to same repository

cheehongw:
### Summary:
Fixes #41 

See discussion in #41 for more info

### Commit Message:

```
Fix uncaught httpError arising from promise

`catchError` does not appear to catch an error on an observable created 
from a promise.

Let's catch the error in the promise before converting it to an 
observable.
```

### DIFF
--- a/src/app/core/services/github.service.ts
+++ b/src/app/core/services/github.service.ts
@@ -590,9 +590,16 @@ export class GithubService {
       headers: { 'If-None-Match': this.issuesCacheManager.getEtagFor(pageNumber) },
       state: 'all'
     });
-    const apiCall$ = from(apiCall);
+    const apiCall$ = from(apiCall.catch(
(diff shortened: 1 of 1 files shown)


Eclipse-Dominator (approved):
LGTM!

kkangs0226:
LGTM, confirmed that the bug has been fixed. Thanks for investigating the`catchError` problem, there could be other parts in the codebase that are causing similar bugs. We can investigate further 👍

chunweii (approved):

Issue #78, opened 2023-02-07
Code on main branch fails Lint tests


vigneshsankariyer1234567890:
Currently, there is quite a lot of code on the main branch that fails Lint tests, probably related to #73 . This does not allow developers who have pre-push hooks enabled to publish their forked branches without fixing these existing bits of code. Let's fix this ASAP first.

PR #79, opened 2023-02-07
Fix lint issues on main

vigneshsankariyer1234567890:
### Summary:
Fixes existing lint issues on main, and #78 . 

### Changes Made:
Changes to parts of code with lint issues.
-

### Commit Message:

```
Let's fix existing lint issues on main.
```


cheehongw (approved):
LGTM! Thanks for fixing these!

PR #82, opened 2023-02-14
Add show and hide labels menu

### LINKED ISSUE #76: Ability to selectively hide certain labels
There might be cases where PR or issues contains a lot of labels and it can easily cluster the current view in the dashboard. It might be wise to include in the ability to selectively hide labels in the UI to tackle such scenarios

Add a dropdown checklist to select labels to remove.

### LINKED ISSUE #80: UI bug for label component
**Describe the bug**

Minor UI bug for label component where after adding multiple labels, the limited component space causes ```labels```( above the search box ) to get pushed up. The mini scroll bar has been scrolled to the top for both screenshots 

**To Reproduce**

1. Go to ```labels``` search bar
2. Add multiple labels 
3. Check that the scroll bar is not able to scroll up to ```labels```

**Expected behavior**
Every part of the label search component should be seen 

**Screenshots**
Screenshot 2023-02-09 at 9 22 57 AM

Screenshot 2023-02-09 at 9 22 44 AM

- Browser Chrome Version 109.0.5414.119 (Official Build) (x86_64)

Eclipse-Dominator:
### Summary:
Adds the ability to show and hide labels in the viewer via an openable menu (closes #76). 
As a side effect of merging with the label filter, this also addresses the UI bug related to the label filter (closes #80)

### Menu demostration
image

### Changes Made:

- Created a new menu to allow users to access and select all current labels
- Merged the functionality of hiding labels with filtering label. 

### Commit Message:

```
The UI can look messy when there are a lot of labels

Let's add a menu to allow the users to selectively hide label from the view
```


src/app/issues-viewer/issues-viewer.component.html
```diff
@@ -49,8 +49,12 @@
       </div>
     </mat-grid-tile>
 
-    <mat-grid-tile class="label-filter-grid-tile" colspan="2">
+    <!-- <mat-grid-tile class="label-filter-grid-tile" colspan="2">
       <app-label-chip-bar [selectedLabels]="this.labelFilter$"></app-label-chip-bar>
+    </mat-grid-tile> -->
+
```
chunweii, line 55: You should remove the commented code

src/app/issues-viewer/label-filter-bar/label-filter-bar.component.ts
```diff
@@ -0,0 +1,110 @@
...
+      },
+      (err) => {},
+      () => {
+        this.initialize();
+      }
+    );
+  }
+
+  private initialize() {
+    this.labelIndex = this.labelService.labels.map((label, index) => index);
```
chunweii, line 91: This just seems like an array of 0 to n-1. Is there a better way of doing this?

chunweii (changes requested):
Overall, good work on implementing the label filter! Please resolve my other comments.

Also, it seems that the original `LabelChipBarComponent` is no longer used. Let's delete the component

src/app/issues-viewer/label-filter-bar/label-filter-bar.component.ts
```diff
@@ -0,0 +1,110 @@
...
+    this.allLabels = this.labelService.labels.map((label) => {
+      return {
+        name: label.getFormattedName(),
+        color: label.color
+      };
+    });
+    this.filteredLabelIndexes = this.labelCtrl.valueChanges.pipe(
+      startWith(''),
```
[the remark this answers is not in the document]
Eclipse-Dominator: initializing `labelCtrl` with initial value of `''` allows `filteredLabelIndexes`  to initialize with the `labelIndex` values instead of starting with an empty list. This prevents the menu to display empty list when first opened by the user

src/app/issues-viewer/label-filter-bar/label-filter-bar.component.ts
Eclipse-Dominator: updated to use `[...this.labelService.labels.keys()]`

gycgabriel (commented):
Thank you for this feature. 

I noticed that after selecting a label filter and hiding that label, the label filter is reset and the hidden label cannot be selected. Is it possible to decouple this behaviour? 

One major use case is I want to filter and hide the label `bug`, since all the filtered issues are of label `bug`

src/app/issues-viewer/card-view/card-view.component.html
```diff
@@ -31,9 +31,15 @@
...
               {{ issue.milestone?.title }}
-              <mat-chip *ngFor="let label of issue.githubLabels" [ngStyle]="labelService.setLabelStyle(label.color)" selected>
-                {{ label.name }}
-              </mat-chip>
+              <ng-container *ngFor="let label of issue.githubLabels">
+                <mat-chip
+                  *ngIf="!issues.dropdownFilter?.hiddenLabels.has(label.name)"
+                  [ngStyle]="labelService.setLabelStyle(label.color)"
+                  selected
+                >
```
kkangs0226, line 39: Just a nitpick, let's not put ```>``` in a new line

src/app/issues-viewer/label-filter-bar/label-filter-bar.component.ts
```diff
@@ -0,0 +1,114 @@
...
+    }
+    this.hiddenLabelNames.delete(label);
+    this.hiddenLabels.next(this.hiddenLabelNames);
+  }
+
+  /**
+   * chip as of the current project version consumes click events
+   * this method is used as an workaround the issue.
+   * https://github.com/angular/components/issues/19759
+   */
```
kkangs0226, line 67: Can I check if this is still an issue for the latest Angular version? If this has been fixed in later versions, maybe we can leave a comment to update this part in the future.

kkangs0226 (commented):
Good work @Eclipse-Dominator 👍 The new label component seems like a much better alternative. 
I've left some comments.

Eclipse-Dominator:
> Thank you for this feature.
> 
> I noticed that after selecting a label filter and hiding that label, the label filter is reset and the hidden label cannot be selected. Is it possible to decouple this behaviour?
> 
> One major use case is I want to filter and hide the label `bug`, since all the filtered issues are of label `bug`

Well, I think that was the original behaviour but I thought it might be a bit wierd since there isn't really a visual UI to indicate a hidden label is selected from the main screen which is why I opt to disable the select after hiding a label. I can decouple it quite easily.

gycgabriel:
There is also a bug: when typing in search labels, the filter is reset. 

You could leave the checkbox checked but disabled to indicate the filter is applied but the label is hidden

Eclipse-Dominator:
> There is also a bug: when typing in search labels, the filter is reset.
> 
> You could leave the checkbox checked but disabled to indicate the filter is applied but the label is hidden

Yup thanks for spotting the bug. The issue is caused by ngModel shrinking the selectedLabelNames when the elements are filtered. I have changed it to use css level hiding instead and removed the dynamic filtering. This should be more efficient compared to before.

src/app/issues-viewer/card-view/card-view.component.html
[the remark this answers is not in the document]
Eclipse-Dominator: I think pretty-quick automatically formatted the html tag like this when I pushed.

src/app/issues-viewer/label-filter-bar/label-filter-bar.component.ts
Eclipse-Dominator: I'm not sure if this is fixed since chip components based on angular material were originally used as a form of a button, hence it's likely that it's the intended behavior to consume click events when it's not disabled.

chunweii (approved):
LGTM. I understand it is difficult to test right now because many of the tests are not valid. But we should still write tests for the components whenever possible.

PR #84, opened 2023-02-19
68 upgrade angular version

Echomo-Xinyu:
### Changes Made:

Upgrade the Angular version from v8.2 to v9

-

### Commit Message:

```
Let's upgrade the Angular version from v8.2 to v9.
```


kkangs0226:
Please resolve conflicts first before we review your PR!

chunweii:
On Windows 11, node v12.22.12, there are some compilation errors due to `ngx-mat-select-search` and other errors in `repo-change-form.component.ts`. I got `src/app/shared/repo-change-form/repo-change-form.component.html:5:39 - error TS2341: Property 'repoName' is private and only accessible within class 'RepoChangeFormComponent'`

kkangs0226:
@CATcher-org/2223s2 Other developers, reminder to help test this PR as well! As @chunweii did, you should specify your environment e.g. browser version

Echomo-Xinyu:
> On Windows 11, node v12.22.12, there are some compilation errors due to `ngx-mat-select-search` and other errors in `repo-change-form.component.ts`. I got `src/app/shared/repo-change-form/repo-change-form.component.html:5:39 - error TS2341: Property 'repoName' is private and only accessible within class 'RepoChangeFormComponent'`

Hi  @chunweii I have fixed the issues by updating the version requirement for `ngx-mat-select-search`.

kkangs0226 (changes requested):
Testing on Mac Monterey 12.6.3, running ```npm start``` gives an error:
image
I think there is a typo.

Echomo-Xinyu:
> I think there is a typo.

Yes, it is indeed a typo. Fix in the latest commit. Thanks for point it out!

kkangs0226:
I am still unable to start the desktop application: 
image
image

Echomo-Xinyu:
> I am still unable to start the desktop application: image image

I have looked into the issue and tried to set `"type":"module"` in `package.json` but it causes many subsequent problems as `require()` is then no longer allowed, despite it has been used extensively in `main.ts`. I have also referred to the CATcher side's PR to [upgrade angular version to 9](https://github.com/CATcher-org/CATcher/pull/955) but there is no similar change appeared in the files changed. Even by now, the `import` and `require()` both exist together with angular version 10 in CATcher. So I suspect there may be some other (and potentially easier) ways to fix the problem.

chunweii:
Are we going to remove electron support for WATcher as well? If it is not our intention to use electron, I think we can safely remove it in a future PR

damithc:
> Are we going to remove electron support for WATcher as well? If it is not our intention to use electron, I think we can safely remove it in a future PR

Yes, we should remove it from WATcher as well.

Echomo-Xinyu:
> The error is caused by using `ES2015` instead of `ES5`. CATcher is still using `ES5`, the easiest way to fix this problem will be to stick to `ES5`.

Thank you for the investigation! I have updated the target version and the current application works fine on my Mac Monterey too.

If no more issues with this upgrade, can you remove the "request for change" and approve the PR?

chunweii (approved):
LGTM. 
Works on Windows 11 22H2.
Tested on
- Chrome Version 111.0.5563.65 (Official Build) (64-bit)
- Firefox Version 111.0 (64-bit)
- Edge Version 111.0.1661.44 (Official build) (64-bit)

kkangs0226 (approved):

kkangs0226:
Since this is a non-trivial PR, we should include important changes in the commit message. Please follow this guideline https://se-education.org/guides/conventions/git.html

PR #92, opened 2023-03-05
Remove label-filter-bar as module export

cheehongw:
### Summary:
Currently, `LabelFilterBarComponent` is declared in both `LabelFilterBarModule` and `IssuesViewerModule`, which causes build errors. 

Let's remove the module `LabelFilterBarModule` as the component is part of `IssuesViewerModule`

The alternative was to keep the module `LabelFilterBarModule`, and add it as an import in `IssuesViewerModule`. But since `LabelFilterBarComponent` is part of `IssuesViewerComponent`, it should not be a module on its own.

### Commit Message:

```
Remove label-filter-bar as module export

Let's remove label-filter-bar as an module as it is not a standalone 
module but a component under the issues-viewer module
```


Eclipse-Dominator:
I think keeping it as part of the issue viewer module is fine. LGTM

Eclipse-Dominator (approved):
LGTM!

chunweii (approved):
LGTM

PR #97, opened 2023-03-18
Improve UIX for milestones and label-filter-bar

### LINKED ISSUE #96: Suggestion to improve UI with search-filter-tag button

Currently, the search-filter-tag button (red button at top right corner) has too strong contrast compared with the other elements in the page and can be distracting to users.

image

We may consider making the button having similar style as the other filter sections and be less contrast against the rest.

### LINKED ISSUE #98: Issues/PRs without milestones missing when all milestones are selected

When no milestones are selected by default, all the issues and PRs will be displayed as below: (especially those highlighted in red box)

Screenshot 2023-03-20 at 12 37 43 PM

But when all the milestones are selected, those without milestones are filtered out (eg those in the red box before disappeared now):

Screenshot 2023-03-20 at 12 38 30 PM

We can have a placeholder to include all those issues and PRs without a milestone.

The current approach where those without milestones will appear when no milestones are selected. However, this design may be unclear to users without any explanation.

**Additional context**
This is related to an ongoing update for the UI for issue viewer #97

### LINKED ISSUE #103: Milestones not sorted when displayed

When the milestones are displayed, they are arranged in arbitrary order from the perspective of users. (eg v1.4 comes before v1.3 as below)

image

The milestones are sorted by their titles: v1.3 comes before v1.4 in the above case.

Echomo-Xinyu:
### Summary:
Fixes #96 
Fixes #98 
Fixes #103 

### Changes Made:

- Update Label Filter Bar color (red to white) and text weight (unbold) to reduce contrast compared with others
- Add default selected values for milestones (all selected)
- Add default milestone for issues/PRs without a milestone
- Add sorting rule for milestones (sort by titles)

Before:

image

After:

image

Screenshot 2023-03-20 at 10 12 40 PM

### Commit Message:

```
Let's update the UIX for milestone and label filter bar:

- the Label Filter Bar has been updated with its background colour and font-weight
- the issues/PRs without milestones have been added a default milestone for filter purpose
- the default milestone values for the filter have been updated to avoid confusion
- the milestones displayed in the filter bar will be sorted automatically to aid searching by users
```


Echomo-Xinyu:
Hi Prof @damithc, this is the change you asked for last time during the meeting. Due to the nature of the label filter bar (the button that gets modified) --  supporting more complicated actions compared with the other dropdown filter bar, it is very hard to get the label filter bar to have a uniform style as the other dropdown filter bar (but still possible). I have removed its colour to reduce the contrast. Do you think it is sufficient for now?

damithc:
@Echomo-Xinyu I see. Does the text need to be bold? Can we show a downward arrow, similar to other filters?

On a separate note, I think we should show '... selected | ... hidden' for milestones as well. In fact it can be `.../... selected` e.g., `4/5 selected`.

e.g,,

```
  Milstones         Labels
3/4 selected      8/8selected
```

Echomo-Xinyu:
> @Echomo-Xinyu I see. Does the text need to be bold? Can we show a downward arrow, similar to other filters?
> 
> On a separate note, I think we should show '... selected | ... hidden' for milestones as well. In fact it can be `.../... selected` e.g., `4/5 selected`.
> 
> e.g,,
> 
> ```
>   Milstones         Labels
> 3/4 selected      8/8selected
> ```

Sure I can work on this direction.

Echomo-Xinyu:
> On a separate note, I think we should show '... selected | ... hidden' for milestones as well. In fact it can be `.../... selected` e.g., `4/5 selected`.
> 
> e.g,,
> 
> ```
>   Milstones         Labels
> 3/4 selected      8/8selected
> ```

Hi Prof @damithc, may I clarify this part?

Right now the design is that: when no milestone is selected, it will appear as this
image
And when we select certain milestones, it will display the selected milestone as here:
image

When many milestones are chosen, the milestones selected are only displayed partially:
image

I personally feel that the current design suits well for CS2103 team project since we typically have 4 - 6 milestones in total (by default all teams should have v1.1, v1.2, v1.3 and v1.4 with some having v1.2b and v1.3b) and we usually either look at overall or simply one or two milestones. In such a use case, I feel it is better to have the milestone displayed clearly rather than showing the number of milestones selected. What do you think?

damithc:
> I personally feel that the current design suits well for CS2103 team project since we typically have 4 - 6 milestones in total (by default all teams should have v1.1, v1.2, v1.3 and v1.4 with some having v1.2b and v1.3b) and we usually either look at overall or simply one or two milestones. In such a use case, I feel it is better to have the milestone displayed clearly rather than showing the number of milestones selected. What do you think?

@Echomo-Xinyu this is fine too, but there is no reason to change behavior before/after selection. When the page is loaded initially, all open milestones should be selected by default and hence the appearance should be as follows from the beginning:
image

There is no reason for it to appear as follows at any point:
image

damithc:
When no milestone is selected, it can be something like
```
Milestones
   -
```
or
```
Milestones
   None
``` 
The point is, the `Milestone` label should stay in the same place/size all the time. There is no good reason to for it to change location at various times.

Echomo-Xinyu:
After some preliminary investigation, I think the rationale for the current milestone design is to make use of the [multi-select in material angular](https://material.angular.io/components/select/overview#multiple-selection) which will display the field title in the options box when there is no option selected.

I think this is the simplest way to handle the multi-select issue here and also makes use of the default option by angular material. But of course, we can also do it in a way similar to the label filter bar. Which way would you prefer then? @damithc

damithc:
> I think this is the simplest way to handle the multi-select issue here and also makes use of the default option by angular material. But of course, we can also do it in a way similar to the label filter bar. Which way would you prefer then? @damithc

In that case we can go with the default behavior. Nevertheless, all open milestones should be selected by default, which means it will look like this from the start:
image

Echomo-Xinyu:
Noted! Let me try to research on how to set default values for the multi-select case.

Echomo-Xinyu:
Hi Prof @damithc, I have added a default milestone with the title "without milestone" for all the issues/PRs without a milestone and added support to filter them through the dropdown filter. Do you have any more suggestions to further polish the feature or regarding the naming convention?

damithc:
> Hi Prof @damithc, I have added a default milestone with the title "without milestone" for all the issues/PRs without a milestone and added support to filter them through the dropdown filter. Do you have any more suggestions to further polish the feature or regarding the naming convention?

@Echomo-Xinyu can give a screenshot?

Echomo-Xinyu:
> @Echomo-Xinyu Looks good. Perhaps use `Without a milestone` or `Not in a milestone`?

Noted! I adopt the "Without a milestone" here.

src/app/core/models/issue.model.ts
```diff
@@ -123,7 +123,8 @@ export class Issue {
     this.title = githubIssue.title;
     this.hiddenDataInDescription = new HiddenData(githubIssue.body);
     this.description = Issue.updateDescription(this.hiddenDataInDescription.originalStringWithoutHiddenData);
-    this.milestone = githubIssue.milestone ? new Milestone(githubIssue.milestone) : null;
+    // githubIssue without milestone will be set to default milestone
+    this.milestone = githubIssue.milestone ? new Milestone(githubIssue.milestone) : Milestone.DefaultMilestone;
```
Eclipse-Dominator, line 127: With this change I think we are safe the assume that the milestone property will always be non-null/defined. Let's remove the optional parameter in line 30 to reflect the changes.
```milestone?: Milestone;``` -> ```milestone: Milestone```

File: src/app/issues-viewer/card-view/card-view.component.html [hunk 2]
```diff
@@ -29,8 +29,8 @@
           </mat-card-header>
           <mat-card-content>
             <mat-chip-list aria-label="Labels">
-              <span *ngIf="issue.milestone" class="octicon-milestone" octicon="milestone" color="grey" size="8"></span>
-              {{ issue.milestone?.title }}
+              <span *ngIf="issue.milestone.state" class="octicon-milestone" octicon="milestone" color="grey" size="8"></span>
+              {{ issue.milestone?.title !== 'Without a milestone' ? issue.milestone?.title : '' }}
```
Eclipse-Dominator, line 33: I think checking with `issue.milestone.state` is more efficient than checking with the string value of title. It is possible for users to intentionally set the milestone title as 'Without a milestone' but not possible for them to set the state to `null`

Eclipse-Dominator (changes requested):
Features works as suggested. Good work!

src/app/issues-viewer/card-view/card-view.component.html
Echomo-Xinyu: Thanks for the suggestion!

Eclipse-Dominator (approved):
LGTM!

kkangs0226 (changes requested):
Seems like this PR is still based on Angular v8, please resolve conflicts first! Using Angular v9, I am running into errors.

Echomo-Xinyu:
> Seems like this PR is still based on Angular v8, please resolve conflicts first! Using Angular v9, I am running into errors.

Fixed!

kkangs0226 (approved):
LGTM, good work!

chunweii:
Before we can merge this PR, please follow the commit message guidelines

Issue #101, opened 2023-03-20
Unable to switch repo when invalid repo is given


vigneshsankariyer1234567890:
**Describe the bug**
When an invalid repo is given, the user is unable to switch to another repository. This could be related to #47 and needs to be handled.
 
**To Reproduce**

1. Run WATcher
2. Look up for an invalid repository (could be 'hello' or even 'hello/hello')
3. Following state occurs:

<img width="1440" alt="image" src="https://user-images.githubusercontent.com/68676914/226254734-5f13f962-eb28-4896-a19e-d67b4a02442d.png">

**Expected behavior**
I should be able to change the repo regardless of my input.

- OS: MacOS 12
- Browser Chrome
- Version node v 14

**Additional context**

chunweii:
This has been fixed in #104.

PR #104, opened 2023-03-20
Better error messages

### LINKED ISSUE #47: Error messages for invalid repo not friendly

vigneshsankariyer1234567890:
### Summary:
Fixes #47 and #101. Previously, a PR was written to solve #47 but it warranted additional investigation and was hence closed.

Originally, if a repo name with the wrong name or organisation was provided, it was difficult and not user friendly to see the what caused the error. Furthermore, certain methods which requested issues or milestones from Github did not always catch the subsequent errors that were thrown. At the same time, a progress bar would be visible within the issues-viewer component that indicated the page was still loading, and there was no option to switch to another repo. 

This PR resolves those issues as both #47 and #101 are linked issues. 

### Changes Made:
- Added more validation in `Repo` to ensure given repos are valid
- Added clearer logging to `GithubService` for better debugging.
- Changed behaviour of PhaseService to validate org and repoName from localStorage
- Edited components that request for repo-based information to validate repos before requesting
- Changed `LabelFilterBarComponent` to complete loading regardless of errors received

### Commit Message:
```
Add better error messages for invalid repo inputs.

Long and unreadable errors are produced when invalid repos are provided in auth.

This makes the app difficult to use and unintuitive.

Catching and validating these errors will help us create a better experience for the user.

Let's
* Add more validation in `Repo` to ensure given repos are indeed valid
* Add clearer logging to `GithubService` for better debugging.
* Change behaviour of PhaseService to validate org and repoName from localStorage
* Edit components that request for repo-based information to validate repos before requesting
* Change `LabelFilterBarComponent` to complete loading regardless of errors received
```


src/app/core/models/repo.model.ts
```diff
@@ -16,13 +16,36 @@ export class Repo {
...
+      throw new Error('Invalid repo name. Please provide repo name in the format Org/Repo.');
     }
     return new Repo(repoUrlSplit[0], repoUrlSplit[1]);
   }
 
+  public static ofEmptyRepo() {
+    return EMPTY_REPO;
+  }
+
+  public static isEmptyRepo(repo: unknown) {
```
kkangs0226, line 28: I think ```isInvalidRepoName``` is more suitable here since it returns false when it is not a repo as well. Also makes more sense when it is called in ```src/app/shared/layout/header.component.ts```

src/app/core/services/github.service.ts
```diff
@@ -379,7 +379,7 @@ export class GithubService {
       map((response) => {
         return response['data'];
       }),
-      catchError((err) => throwError('Failed to fetch assignable users for repo.'))
+      catchError((err) => throwError('Failed to fetch assignable users for repo (getUsersAssignable).'))
```
kkangs0226, line 382: Let's avoid exposing the method name in the error message we show to users, same for the rest of the error messages

src/app/issues-viewer/issues-viewer.component.ts
```diff
@@ -120,4 +124,21 @@ export class IssuesViewerComponent implements OnInit, AfterViewInit, OnDestroy {
...
+   * Checks if our current repository available on phase service is indeed a valid repository
+   */
+  private checkIfValidRepository() {
+    const currentRepo = this.phaseService.currentRepo;
+
+    if (Repo.isEmptyRepo(currentRepo)) {
+      throw new Error('Invalid repo name. Please provide repo name in the format Org/Repo.');
+    }
+
+    this.phaseService.isValidRepository(currentRepo).subscribe((isValidRepository) => {
```
kkangs0226, line 138: Is there a reason for calling the ```isValidRepository``` method in ```phaseService``` rather than calling ```this.githubService.isRepositoryPresent(repo.owner, repo.name)``` directly?

src/app/issues-viewer/issues-viewer.component.ts
```diff
@@ -120,4 +124,21 @@ export class IssuesViewerComponent implements OnInit, AfterViewInit, OnDestroy {
...
+      if (!isValidRepository) {
+        throw new Error('Invalid repo name. Please check your organisation and repo name.');
```
kkangs0226, line 140: Nitpick: For error messages shown to users, let's use ```repository```

kkangs0226 (commented):
Thanks for the detailed summary of your PR and good work! 😀

I've left some comments for minor things I want to check with you.

Another thing to note, for your commit message it does not have to state every implementation change, you can refer to the https://se-education.org/guides/conventions/git.html

HOW the change was done

src/app/auth/confirm-login/confirm-login.component.ts
```diff
@@ -58,9 +60,24 @@ export class ConfirmLoginComponent implements OnInit {
...
+            return of(false);
+          }
+          return this.phaseService.isValidRepository(currentRepo);
+        }),
+        flatMap((isValidRepository) => {
+          if (!isValidRepository) {
+            return new Observable();
+          }
+          return this.githubEventService.setLatestChangeEvent();
+        })
```
chunweii, line 79: `flatMap` is deprecated. Please use `mergeMap`. See https://rxjs.dev/api/operators/flatMap

src/app/issues-viewer/issues-viewer.component.ts
vigneshsankariyer1234567890: It was to keep it consistent with the rest of the changes in the other components. In confirm-login, it didn't have an existing dependency to githubService while some of the other components had it. I thought it was better to then use phase service instead.

src/app/core/services/phase.service.ts
```diff
@@ -98,7 +102,16 @@ export class PhaseService {
    * Retrieves the repository url from local storage and sets to current repository.
    */
   initializeCurrentRepository() {
-    const repo = new Repo(window.localStorage.getItem('org'), window.localStorage.getItem('dataRepo'));
+    const org = window.localStorage.getItem('org');
+    const repoName = window.localStorage.getItem('dataRepo');
+    this.logger.info(`Phase Service: received initial org (${org}) and initial name (${repoName})`);
+    let repo: Repo;
+    if (org === 'undefined' || repoName === 'undefined') {
```
chunweii, line 109: I do not think that you should check against the string `'undefined'`. You should check using `!org` and `!repoName`

chunweii (changes requested):
Thank you for the PR. ~~Other than my comment for phase service, please try writing some tests.~~ Tests are difficult to write now, so we can defer this for now.

src/app/core/services/phase.service.ts
vigneshsankariyer1234567890: Originally I had a check for this to see if `org` or `repoName` were truthy. However, localStorage's implementation is pretty funky. In `session-selection.component` when `org` and `repoName` are initially set, if the value itself is undefined or even null, `window.localStorage::setItem` essentially sets a key-value pair of the key to the string value of the the item being set. So, even if `org` or `repoName` were indeed undefined, they would be be set to the string `'undefined'`, which is why the check is required. Pretty weird IMO.

vigneshsankariyer1234567890: I think what we can do instead is probably do a `JSON::parse` on the item that is retrieved from localStorage, I'll try and see if that helps to convert the value to the actual `undefined` value instead of a string of `'undefined'`

vigneshsankariyer1234567890: that didn't work either as `JSON::parse` now throws an error if there is no `[` or `{` at the start of the string

src/app/issues-viewer/issues-viewer.component.ts
```diff
@@ -119,4 +123,21 @@ export class IssuesViewerComponent implements OnInit, AfterViewInit, OnDestroy {
...
   }
+
+  /**
+   * Checks if our current repository available on phase service is indeed a valid repository
+   */
+  private checkIfValidRepository() {
+    const currentRepo = this.phaseService.currentRepo;
+
+    if (Repo.isInvalidRepoName(currentRepo)) {
+      throw new Error('Invalid repository name. Please provide repo name in the format Org/Repository.');
```
kkangs0226, line 134: repo -> repository

src/app/core/services/github.service.ts
```diff
@@ -457,7 +457,7 @@ export class GithubService {
       map((response) => {
         return response['data'];
       }),
-      catchError((err) => throwError('Failed to fetch events for repo.'))
+      catchError((err) => throwError('Failed to fetch issue events for repo'))
```
kkangs0226, line 460: repo -> repository

src/app/core/services/github.service.ts
```diff
@@ -379,7 +379,7 @@ export class GithubService {
       map((response) => {
         return response['data'];
       }),
-      catchError((err) => throwError('Failed to fetch assignable users for repo.'))
+      catchError((err) => throwError('Failed to fetch assignable users for repo'))
```
kkangs0226, line 382: repo -> repository

src/app/core/services/phase.service.ts
vigneshsankariyer1234567890: @chunweii what if in `SessionSelectionComponent`, when calling `SessionSelectionComponent::setupSession`, we check whether the `repoOrg` or `repoName` is undefined before setting it in localStorage? Currently, we set the values in localStorage regardless of whether it is undefined or not, which localStorage is simply setting as a string. That way, when we check again for org or repo in `phaseService`, if it was not set it would return null. Then we can check and verify that either values are truthy. WDYT?

chunweii: @vigneshsankariyer1234567890 Yes I think that is a good idea

src/app/auth/session-selection/session-selection.component.ts
```diff
@@ -56,9 +56,15 @@ export class SessionSelectionComponent implements OnInit {
...
+     * Since localStorage::setItem with an undefined value can result in
+     * the subsequent value being stored as a string being 'undefined', check
+     * if undefined before storing it.
      */
-    window.localStorage.setItem('org', repoOrg);
-    window.localStorage.setItem('dataRepo', repoName);
+    if (repoOrg && repoName) {
+      window.localStorage.setItem('org', repoOrg);
+      window.localStorage.setItem('dataRepo', repoName);
+    }
```
chunweii, line 67: For the `else` case, we should clear the `localStorage` for `'org'` and `'dataRepo'` so that the values from the old sessions will not be used.

chunweii (changes requested):
One last nitpick, otherwise, LGTM

src/app/auth/session-selection/session-selection.component.ts
```diff
@@ -64,6 +64,8 @@ export class SessionSelectionComponent implements OnInit {
     if (repoOrg && repoName) {
       window.localStorage.setItem('org', repoOrg);
       window.localStorage.setItem('dataRepo', repoName);
+    } else {
+      window.localStorage.clear();
```
chunweii, line 68: I'm sorry if I wasn't clear in my previous comment, but you should not use `clear()` for the whole local storage. Instead, only delete the items `org` and `dataRepo`. There are other things that are kept in the local storage that should be kept, including the logs.

vigneshsankariyer1234567890: oh, i see haha. i'll make the relevant changes!

vigneshsankariyer1234567890: maybe we should remove the items first before setting it? might be better than an else remove

chunweii (approved):

src/app/core/models/repo.model.ts
```diff
@@ -16,13 +16,36 @@ export class Repo {
   public static of(repoUrl: string) {
     const repoUrlSplit = repoUrl.split('/');
     if (repoUrlSplit.length !== 2) {
-      return undefined; // throw error?
+      throw new Error('Invalid repository name. Please provide repo name in the format Org/Repository Name.');
```
kkangs0226, line 19: repo -> repository

Issue #110, opened 2023-03-25
Issue not shown in dashboard


Echomo-Xinyu:
**Describe the bug**

Currently, there are two issues created by "Echomo-Xinyu" in the repository ["CATcher-org/WATcher-test"](https://github.com/CATcher-org/WATcher-test/issues?q=is%3Aissue)

<img width="1640" alt="image" src="https://user-images.githubusercontent.com/25452497/227712941-260db8e8-88d1-437e-83d5-1b43d23646fd.png">

However, the issues cannot be seen from the WATcher dashboard, despite the type selected is All.

<img width="1920" alt="image" src="https://user-images.githubusercontent.com/25452497/227713075-04f31c2f-80ee-4f5e-aa3d-2ebdfc3cba82.png">

**To Reproduce**

1. Go to WATcher and watch the "CATcher-org/WATcher-test"

**Expected behavior**
The issues should be shown clearly as well.

**Desktop (please complete the following information):**

- OS: Mac Monterey
- Browser: Chrome
- Version: Version 111.0.5563.64 (Official Build) (arm64)

vigneshsankariyer1234567890:
Could this be related to #107?

Echomo-Xinyu:
> Could this be related to #107?

Yes, I think they should be the same kind of bugs, at least appear to be the same.

Eclipse-Dominator:
Issues fixed by #118

PR #118, opened 2023-04-03
Add additional card to show unassigned issues

### LINKED ISSUE #107: Issues not showing for some repositories
**Describe the bug**
For some repositories, the issues/PRs are not displayed.

**To Reproduce**
Try to access `chunweii/bugreporting` which has 14 issues, and 1 milestone. It shows:
image
The same thing happens for `kkangs0226/bugreporting` which has 9 issues and no milestones.

vigneshsankariyer1234567890:
### Summary:
Fixes #107 and #110.

Originally, IssuesViewerComponent would produce a set of CardViewComponents by iterating through the list of assignees for that repository and passing each assignee to the CardViewComponent. CardViewComponent in turn would create its own IssueDataTable, which would produce a list of issues, filtered by whether that issue was indeed assigned to the assignee passed to the IssueDataTable. However, not all repositories would have issues with assigned users. In fact, it is possible for issues to not have any assigned users at all. This leads to certain issues not being shown under anyone, which can be quite misleading. 

Instead, let's add a CardViewComponent that contains only unassigned issues. This will lead to a better UX as all issues are now visible. 

### Changes Made:
- Changed `IssuesDataTable` to filter for unassigned issues if no assignee was passed
- Added an additional CardViewComponent after list of CardViewComponents iterated by assignee

### Commit Message:
```
Add an additional CardViewComponent that displays unassigned Issues.
```


src/app/shared/issue-tables/IssuesDataTable.ts
```diff
@@ -105,6 +105,11 @@ export class IssuesDataTable extends DataSource<Issue> {
                     return issue.assignees.includes(this.assignee.login);
                   }
                 });
+              } else {
+                // If no assignee, filter the unassigned issues only.
+                data = data.filter((issue) => {
+                  return issue.assignees.length === 0;
```
Echomo-Xinyu, line 111: this should be `return issue.issueOrPr !== "PullRequest" && issue.assignees.length === 0;` to avoid the problem of PR appearing twice (refer to overall comment for the screenshot)

Echomo-Xinyu (changes requested):
Thanks for the investigation and quick fix man! @vigneshsankariyer1234567890 

Screenshot 2023-04-03 at 11 46 34 PM

Currently, the PR highlighted in red box appears twice when testing with the repo "AY2223S1-CS2103T-T09-4/tp" and the fix is suggested.

Meanwhile, I think we should aim to make the size of the unassignee's mat-card-view the same as those with a profile photo.

Echomo-Xinyu:
> In fact, should we even show unassigned PRs in the 'Unassigned Issues' column? For PRs, what matters is the author and not the assignees.

No, I think PR should only be grouped by its author not the assignee, which is also the current way of working. I have suggested the fix such that only issues will appear in the "Unassigned Issues" column.

src/app/shared/issue-tables/IssuesDataTable.ts
vigneshsankariyer1234567890: thanks for the catch, just added a commit adding your change :)

Echomo-Xinyu (approved):
Tested and LGTM!

kkangs0226:
Actually, I think it might be better to show issues under the user who created the issue, not the user assigned to that issue. 

Usually the user who has been assigned to the issue will raise a PR to fix the issue, which will be reflected under that user on WATcher as well. So there will be 1 issue and 1 PR for the same problem listed under that person, which is quite redundant. Also, the user who investigated the issue & reported the issue is not getting any credit once the issue gets assigned to another user.

TLDR: instead of adding additional card to show unassigned issues, we could just list any issue under the user who created it. 

@CATcher-org/2223s2 What are your thoughts?

damithc:
> Actually, I think it might be better to show issues under the user who created the issue, not the user assigned to that issue.

@kkangs0226 in CS2103 context, we are more interested in the issue assignee than the creator. That is, "who is allocated to do which task?"

chunweii:
> TLDR: instead of adding additional card to show unassigned issues, we could just list any issue under the user who created it.

Perhaps we can add the information about the creator of the issue to each issue card? Hopefully that does not clutter the card too much..

damithc:
My current thinking is, in this dashboard, let's not bother about the creator of the issue. Reason: This dashboard is meant to monitor work distribution and progress. Issue creator doesn't play a significant role in that objective. Therefore, it's cleaner to completely ignore the issue creator in this dashboard. But it makes sense to have a separate column that shows unallocated work (i.e., unassigned issues) as that is directly related to work allocation.

Work related to issue tracker maintenance: e.g., issue creation, triaging, reviewing, allocating, etc. can appear in a separate dashboard that we might implement in the future.

What do you guys think?

Eclipse-Dominator:
Perhaps it might be better 2 have 2 different views to view creation and assignment since mixing both here is quite confusing. 

Perhaps something like this?
image

There is also argument for PRs as whether to display assigned or created.

chunweii:
> Perhaps it might be better 2 have 2 different views to view creation and assignment since mixing both here is quite confusing.

Sounds like a good idea. We might want to add a description to the toggle so that the user will know what the toggle does.

Let's work on this on a separate PR. You can open an issue for this

kkangs0226:
As for now, leaving ```unassigned users``` on the same page seems OK to me, since it can be seen as a separate user column for "no users". Moving this column to a separate toggle view might not be consistent with existing page design, since we would only have one column for that page. 

We can move on with the current implementation first and discuss the toggle view in another PR as @chunweii has mentioned 👍

kkangs0226 (commented):
LGTM, we can merge this after you fix the lint errors

damithc:
The purpose of this particular dashboard is to show work allocation and progress. It will answer questions such as these:
1. Who is supposed to do what (for this milestone)?
2. What's the progress of each person's allocated work? How many items are done? How many are in progress? How many are not even started yet?
3. X is falling behind. Who has less work pending and hence take over X's work? 
4. Y is done with her work early. What else she can do?

So, it makes sense to have one column for each dev showing what they have been allocated, and another column showing what work still remains unallocated. It will help with the 4th question above.

This is also why showing assignees of PRs doesn't make sense in this view. For one thing, PR assignment is not part of the CS2103/T workflow. For another, we are not sure how PR assignment affects work allocation.

Something missing from this view is the information on PR review allocation. As 'reviewing a PR' is work too, it would be nice if that can be shown in this view. i.e., in addition to issues assigned, and PRs authored, also show PRs assigned for review. OK if it is hard to do though.

vigneshsankariyer1234567890:
@damithc Understood, prof! I think it's doable so let's explore that as a separate issue!

PR #120, opened 2023-04-04
119 fix datatable inefficiency

Eclipse-Dominator:
### Summary:
Currently, a new subscriber is being added to the Observables found in `IssueDataTable` class each time a new issue is being retrieved from `issueService`. This creates at least an n^2 additional runtime which consists of completely unnecessary yet expensive function calls to update the view based on the filter stack. When the method should really be only called once.

### Changes Made:
- Remove unnecessary nesting of subscribers in `IssueDataTable`

### Commit Message:

```
Inefficient nesting of subscribers in IssueDataTable has introduced 
unnecessary and very expensive computes to WATcher.

Let's 
- Improve WATcher's efficiency by removing the nesting of subscribers in IssueDataTable.ts
```

### DIFF
--- a/src/app/shared/issue-tables/IssuesDataTable.ts
+++ b/src/app/shared/issue-tables/IssuesDataTable.ts
@@ -2,7 +2,7 @@ import { DataSource } from '@angular/cdk/table';
 import { MatPaginator } from '@angular/material/paginator';
 import { MatSort } from '@angular/material/sort';
 import { BehaviorSubject, merge, Observable, Subscription } from 'rxjs';
-import { flatMap, map } from 'rxjs/operators';
+import { map } from 'rxjs/operators';
 import { GithubUser } from '../../core/models/github-user.model';
 import { Issue } from '../../core/models/issue.model';
 import { IssueService } from '../../core/services/issue.service';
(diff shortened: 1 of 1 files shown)


chunweii:
I am getting this error message when switching repositories:

```
TypeError: Cannot read properties of undefined (reading 'unsubscribe')\n    at IssuesDataTable.disconnect
```

Can you check the `disconnect()` method and make sure the subscriptions are unsubscribed correctly?

kkangs0226 (approved):
LGTM, confirmed that the number of function calls is back to normal. 👍

vigneshsankariyer1234567890 (approved):
Good catch, new changes are working fine on my side as well. LGTM 🚀

Eclipse-Dominator:
> Thank you for the PR. The changes indeed speed up the loading time significantly. I can now load the CATcher repository :)
> 
> But...
> 
> I am getting this error message when switching repositories:
> 
> ```
> TypeError: Cannot read properties of undefined (reading 'unsubscribe')\n    at IssuesDataTable.disconnect
> ```
> 
> Can you check the `disconnect()` method and make sure the subscriptions are unsubscribed correctly?

Right I forgot to set the issueSubscription subscription, I added it in CATcher but forgot to include it in WATcher. Good catch!

chunweii (approved):

PR #122, opened 2023-04-05
Add horizontal scroll to Issues Viewer

### LINKED ISSUE #52: UI Improvements for when there are too many users
image

The list of users extend beyond the header

vigneshsankariyer1234567890:
### Summary:
Resolves #52 by adding a horizontal scroll that ensures that cards do not extend beyond the existing width of page.

Originally, if screen width was decreased or if a team had more than 4-5 members, there would be `CardViewComponents` beyond the page. This leads to a dampened UX

### Changes Made:
- Added horizontal scrolling for `IssuesViewerComponent`

### Commit Message:

```
Let's add horizontal scrolling for `IssuesViewerComponent`
```


Eclipse-Dominator (changes requested):
Nice fix! However, when loading in repos with a lot of users, the names of the users tend to overflow and get squished along with the horizontal scrollbar in each issues shown below.
 
image

Since in the code of `CardView`, there is already support for truncating of text
```typescript
<mat-card-title>
              <span class="octicon" [octicon]="this.getOcticon(issue)" [color]="this.getIssueOpenOrCloseColor(issue)"></span>
              #{{ issue.id }}: {{ this.fitTitleText(issue.title) }}
            </mat-card-title>
```
it might be better if we can keep each dialog boxes as the same size as before now that we have a horizontal scrollbar or change the truncating width to prevent the above issue from happening

damithc:
@vigneshsankariyer1234567890  Just to be clear, you are making the entire page horizontally scrollable, right? i.e., not making individual cards scrollable.

Eclipse-Dominator:
> @vigneshsankariyer1234567890 Just to be clear, you are making the entire page horizontally scrollable, right? i.e., not making individual cards scrollable.

I think the individual cards becoming scrollable is a side effect of changing the CSS of the parent components since the changes here did not involve changes in `CardView` Component.

damithc:
> @damithc, I think the individual cards becoming scrollable is an unintended side effect of changing the CSS of the parent components since the changes here did not involve changes in `CardView` Component.

I see. Cards should not be horizontally scrollable. They should show the full title, as before.

kkangs0226 (approved):
LGTM, let's merge this in after the lint errors are fixed.

vigneshsankariyer1234567890:
@Eclipse-Dominator @kkangs0226 @chunweii just bumping again to re-review!

PR #129, opened 2023-04-10
Modularization and abstraction of filter related components

Eclipse-Dominator:
### Summary:
Partially addresses #69 

Profile page requires reusing code from filter-related components. 
To avoid too much code duplication, let's modularise the filter bar and abstract out important aspects of filter-related components

### Changes Made:

- Modularizes Filter bar into its own components
- Moves `labelfilterbar` to a component within the abstracted out `filterbar`
- Abstract out `dropdownfilter` into its own class
- Abstract out interfaces for filterableSource and filterableComponents for better code quality
- Further abstracts out filter functionalities to reduce code duplication

### Commit Message:

```
Profile page requires reusing code from filter-related components. 
To avoid too much code duplication, let's modularise the filter bar and abstract out important aspects of filter-related components

- Modularizes Filter bar into its own components
- Moves labelfilterbar to a component within the abstracted out filterbar
- Abstract out dropdownfilter into its own class
- Abstract out interfaces for filterableSource and filterableComponents for better code quality
- Further abstract out filter functionalities to reduce code duplication
```

Issue #69, opened 2023-02-02
Detailed dashboard view of individual users.


Eclipse-Dominator:
**Is your feature request related to a problem? Please describe.**
Sometimes, we might want to have a detailed view of contributions per user the user can click on a specific contributor of a repository to access a detailed report view of what he or she has done in the past. It is also a nice place to build other ways to view user contribution to a project (i.e. calendar view etc.) We can create individual modules for each different view and have a dashboard-like page for each user accessible via clicking on the username in the WATcher home screen. This also give us the flexibi

**Describe the solution you'd like**
Create specific page to view details of an individual similar to the Github profile page and populate it with information on the user related to the current repository as well as other statistics. 
![image](https://user-images.githubusercontent.com/4567895/216236809-e4affc4e-693b-4b53-8b39-b7e1fe31f5b0.png)

PR #130, opened 2023-04-10
Details page redirection

Eclipse-Dominator:
### Summary:
Partially addresses #69, depends on #129 

Adds basic redirection to the profile page via clicking on the user icon
Currently, issues reset during the initialization of `issue viewer`, we should only initialize it upon repo change to cache old data and not reload every time the user navigates back to the dashboard.

### Changes Made:
- Adds profile page routing to the main page
- Moves page resets to upon repo change
- Adds routing to the profile page
- Enables users to access the profile page by clicking on their profile

### Commit Message:

```
We need to access a new page for the profile page.
Currently, issues reset during the initialization of `issue viewer`, we should only initialize it upon repo change to cache old data and not reload every time the user navigates back to the dashboard.
Let's 
- Adds profile page routing to the main page
- Moves page resets to upon repo change
- Adds routing to the profile page
- Enables users to access the profile page by clicking on their profile
```


src/app/detailed-viewer/detailed-viewer.component.ts
```diff
@@ -0,0 +1,72 @@
...
+    private logger: LoggingService
+  ) {
+    this.repoChangeSubscription = this.phaseService.repoChanged$.subscribe((newRepo) => {
+      this.initialize();
+    });
+  }
+
+  ngOnInit() {
+    if (this.route.snapshot.paramMap.get('name') === undefined) {
+      this.logger.info('detailed-viewer: Missing username');
```
chunweii, line 45:
```suggestion
      this.logger.info('DetailedViewerComponent: Missing username');
```

src/app/detailed-viewer/detailed-viewer.component.ts
```diff
@@ -0,0 +1,72 @@
...
+  styleUrls: ['./detailed-viewer.component.css']
+})
+export class DetailedViewerComponent implements OnInit, OnDestroy, AfterViewInit {
+  public user: GithubUser = undefined;
+
+  private userSubscription: Subscription;
+
+  /** Observes for any change in repo*/
+  repoChangeSubscription: Subscription;
+  issueChangeSubscription: Subscription;
```
chunweii, line 28: This variable seems unused.

src/app/detailed-viewer/detailed-viewer.component.ts
```diff
@@ -0,0 +1,72 @@
...
+    private router: Router,
+    private logger: LoggingService
+  ) {
+    this.repoChangeSubscription = this.phaseService.repoChanged$.subscribe((newRepo) => {
+      this.initialize();
+    });
+  }
+
+  ngOnInit() {
+    if (this.route.snapshot.paramMap.get('name') === undefined) {
```
chunweii, line 44: The return result could be null, which is different from undefined.

Perhaps we can use the falsy value

src/app/issues-viewer/card-view/card-view.component.ts
```diff
@@ -50,4 +52,12 @@ export class CardViewComponent implements OnInit, AfterViewInit, OnDestroy, Filt
   retrieveFilterable(): FilterableSource {
     return this.issues;
   }
+
+  viewUserInBrowser() {
```
chunweii, line 56: Can we rename this to `switchToUserDetailPhase` or something similar?

I was confused by this method initially. I thought this method will open the user's github page in a new tab or browser window.

chunweii (changes requested):
`DetailedViewer` sounds vague. Perhaps we can change the module and component name to `UserDetailViewer`.

Also, you should add a new phase `UserDetailPhase`.

src/app/detailed-viewer/detailed-viewer.component.ts
Eclipse-Dominator: Yes, thanks for pointing out, I misread the return type of paremMap.get()
Luckily a separate check was done before this so this is actually never reached, it's more like defensive programming here.

Eclipse-Dominator:
I have added the new phase as well as setting phase values in the constructor of each page to account for forward and backward navigation of pages

chunweii (approved):

PR #133, opened 2023-04-10
Detail page commit list diffstat

Eclipse-Dominator:
### Summary:
Partially addresses #69, depends on #132 

### Changes Made:
- Adds diffstats component to display line additions and deletions
- Adds RSQ data structures to enable fast RSQ queries
- Calculates cumulative stats in the commits
- Adds the ability to check cumulative stats across multiple commits
- Show currently shown commits

### Commit Message:

```
The commit list currently doesn't display line changes
Let's display the line change per commit as well as the cumulative line changes

- Adds diffstats component to display line additions and deletions
- Adds RSQ data structures to enable fast RSQ queries
- Calculates cumulative stats in the commits
- Adds the ability to check cumulative stats across multiple commits
- Show currently shown commits
```

PR #134, opened 2023-04-10
Detail page commit list calendar range selection

Eclipse-Dominator:
### Summary:
Partially addresses #69, depends on #133 

Currently, there is no way for the user to change the date range of the commit list.

### Changes Made:
- Adds the ability for users to change the current range selection for the commit list
- Adds @angular/material-moment-adapter to enable datepicker to support moment datetimes
- Adds custom dialog for users to select date range
- Adds button for users to change current commit range

### Commit Message:

```
Currently, there is no way for the user to change the date range of the commit list
Let's
- Adds the ability for users to change the current range selection for the commit list
- Adds @angular/material-moment-adapter to enable datepicker to support moment datetimes
- Adds custom dialog for users to select date range
- Adds button for users to change current commit range
```


gycgabriel (approved):
LGTM

gycgabriel:
> The code was written back when we were using an older version of angular material, so there wasn't an option available for date range selection like in [here](https://material.angular.io/components/datepicker/overview#date-range-selection).
> 
> Should we switch over to use the newer date range select or do that in a later PR?

Let's do that in another PR.

PR #139, opened 2023-06-19
Fix broken peer dependencies

Eclipse-Dominator:
### Summary:
- Resolves some missing peer dependencies which would eventually cause npm7+. Currently, these missing peer dependencies only throw warnings on npm6:
image

Tested on Node 14 and Node 16.

### Changes Made:
- Bump `ngx-markdown` from 9.1.1 to 10.1.1
  -  As we have already moved on to Angular 10, it is [recommended ](https://github.com/jfcere/ngx-markdown/releases/tag/v10.0.0)that we upgrade this library to v10.

- Bump 'rxjs' from 6.5.3 to 6.5.5
 - Required after upgrading `ngx-markdown` to 10.1.1

- Bump `codelyzer` from 5.1.2 to 6.0.2
  - codelyzer@5.1.2 required Angular 9 as a peer dependency. Since we have upgraded to Angular 10, we should bump codelyzer to a version that accepts Angular v10 as peer dependency.

- Bump `jasmine-core` from ~3.5.0 to ~3.8.0
  - `karma-jasmine-html-reporter` released minor version 1.7.0, which [dropped ](https://github.com/angular/angular-cli/issues/21326#issuecomment-877576859) [support ](https://github.com/dfederm/karma-jasmine-html-reporter/releases/tag/v1.7.0)for jasmine-core@3.5.0, causing a broken peer dependency. Let's bump jasmine-core up to ~3.8.0 to meet the new requirement.
  - You can see `jasmine-core`'s changelog from 3.50 to 3.80 [here](https://github.com/jasmine/jasmine/blob/main/release_notes/3.6.0.md)

### Proposed Commit Message:
```
Fix broken peer dependencies

Let's bump some of our libraries to a more updated version to resolve 
missing peer dependencies.
```


chunweii (approved):
LGTM.

Side note: Seems like ngx-markdown isn't used in WATcher at the moment. We should be able to remove it to have fewer dependencies

Eclipse-Dominator:
> Member

I think this should be left out in a separate PR since it will also affect some other files

PR #141, opened 2023-06-23
Update website title after changing repository

### LINKED ISSUE #108: Website title does not update after changing repository
On the first log in, the title of the website is changed to `WATcher x.x.x - org/dataRepo`. However, if the repository is changed after signing in, the title of the website does not change.

Let's update the title whenever we change repositories.

chia-yh:
### Summary:

Fixes #108 

### Changes Made:

-`AuthService::setTitleWithPhaseDetail()` is called when changing repositories

### Commit Message:

```
Update website title after changing repository

Website title does not change if repository changes after signing in.

Let's update the website title when changing repositories.
```


Eclipse-Dominator (approved):
Works as intended. LGTM!

PR #142, opened 2023-06-23
Fix pressing back to cause issues to being cleared

### LINKED ISSUE #128: Pressing 'back' clears all issues
## Summary

Pressing 'back' after starting a session reloads same repository, but does not show any issues. 

## Steps to reproduce: 
1. Log into a session 
2. Press back on your browser
3. Same repository, however the issues do not load at all. 

Likely related to #65

Eclipse-Dominator:
### Summary:

Fixes #128 

### Cause of issue
The resetting of `issues` in `issue.service` is caused by the resetting of issues each called when the page is reloaded. (pressing back causes `initialize` to be called which results in the clearing of all issues)

``` typescript
  private initialize() {
    this.checkIfValidRepository().subscribe((isValidRepository) => {
      if (!isValidRepository) {
        throw new Error('Invalid repository name. Please provide repository name in the format Org/Repository.');
      }
    });

    // Fetch assignees
    this.assignees = [];

    this.githubService.getUsersAssignable().subscribe((x) => (this.assignees = x));

    // Fetch issues
    this.issueService.reset(false); // <-----
    this.issueService.reloadAllIssues();
  }
```

Thus, the temporary fix is to move the resetting of issues under the observable responsible when the repository is changed. 

### Potential future changes

However, during the investigation, some extra notes occurred which can lead to a separate issue.

When `fetchIssuesGraphql` from `github.service` is called, an additional call to check whether or not the github issues have been changed is also called via `toFetchIssues` shown below:

``` typescript
  /**
   * Fetches an array of filtered GitHubIssues using GraphQL query.
   * In WATcher, this includes pull requests.
   *
   * @param issuesFilter - The issue filter.
   * @returns An observable array of filtered GithubIssues
   */
  fetchIssuesGraphql(issuesFilter: RestGithubIssueFilter): Observable<Array<GithubIssue>> {
    const graphqlFilter = issuesFilter.convertToGraphqlFilter();
    /*
     * Github Issues consists of issues and pull requests in WATcher.
     */
    const issueObs = this.toFetchIssues(issuesFilter).pipe(
        //...
      })
    );
    const prObs = this.toFetchIssues(issuesFilter).pipe(
      filter((toFetch) => toFetch),
      // ....
    );
    // ...
  }
```

`toFetchIssues` verify if there are any changes using the Github API calls and the result is cached in `IssuesCacheManager`.
This means that there are always 2 copies of issues stored in the application: one obtained from Github API and cached in the cache manager and another obtained using Github GraphQL API which is stored in the `issues`  variable. 

### Changes Made:
- Move resetting of issues to when repo is changed instead of initialize in `IssueViewerComponent`

### Commit Message:

```
Fix pressing back to cause issues to be cleared

Issues stored in issue service are being reset each time initialize is
called in IssuesViewerComponent. This is undesirable as this method
can be called when users manually presses back on the browser. 

As the issues cache is being managed via a cache manager, this will
create a scenario where the cache manager claims that there are no
changes to the repo but the cache has already been cleared. This causes
issues to be cleared and no new issues are being loaded from the GitHub
Server.

Let's resolve this by moving the reset to be called when repo change
event occurs instead.
```


chunweii (approved):
Excellent work in resolving this issue!

chunweii:
> Potential future changes
However, during the investigation, some extra notes occurred which can lead to a separate issue.

> When fetchIssuesGraphql from github.service is called, an additional call to check whether or not the github issues have been changed is also called via toFetchIssues shown below:

Good catch! We should not need to have too many calls to the github API to get the issues. Let's put this as a separate issue for now!

PR #143, opened 2023-06-23
[#72] Remove code related to creating and updating repos

Eclipse-Dominator:
### Summary:
fixes part related to creating/updating repos in #72 

### Changes Made:
- Removes code related to creating and modifying content on GitHub
- Removes unused filters

### Commit Message:

```
As a majority of the WATcher code base is ported over from CATcher,
it includes a lot of unused code. One such area is the code related to 
creating and updating issues on GitHub. Since WATcher is mainly used
to read data from GitHub and not modify data for now. 

Let's remove code that is related to modifying or creating repos
```

Issue #72, opened 2023-02-05
Remove unused code from CATcher


chunweii:
Since WATcher is a fork from CATcher, there are still many instances of old and unused code from CATcher. Let's remove these codes.

- `DataService` (used to access data.csv in CATcher)
- `src/app/core/services/session-fix-confirmation` (component to ask user for permission to create a new repository)
- `fetchSettingsFile()` and `fetchDataFile()` in `github.service.ts`
- All tests related to the above

There could be more unused code in the project. These are some that I have found

vigneshsankariyer1234567890:
Adding some more after my personal investigation
- All files in `src/app/core/models/conflict`
- All files in `src/app/core/models/generators`
- Seems like a lot of models in `src/app/core/models/github` are used by unused methods in `issue.service.ts`
- and many more...

Perhaps we can break up the task into sub-tasks by folders? That way the tasks are much more easier to complete, rather than having to go through all the folders within one issue. Wdyt @chunweii ?

chunweii:
Yes, we should break fix this issue in several PRs. Removing all unused code in 1 PR might be too big of a change.

chia-yh:
I noticed `profileForm` in `session-selection.component.ts` doesn't appear to be used currently, are there any plans on utilising `Profile`s in WATcher in the future? There are also some methods associated with it (e.g. `initProfileForm`, `onProfileSelect`) that could be candidates for removal in the case that there are no future plans for using `Profile` with WATcher.

PR #145, opened 2023-06-23
[#72] Remove package ngx-markdown

Eclipse-Dominator:
### Summary:
Part of the solution to #72 
Follow up after #143 (Review this only after this PR is merged)

### Changes Made:
- Removes ngx-markdown package and the related classes

### Commit Message:

```
ngx-markdown is a markdown formatter/parser that was ported over from
CATcher. As WATcher does not use any form of markdown parser, this
package currently does not serve any use in WATcher.

Therefore, let's remove the unused package and the related components.
```


gycgabriel (approved):
LGTM

Issue #74, opened 2023-02-06
Keep User Profiles hovering after scrolling down


Echomo-Xinyu:
**Is your feature request related to a problem? Please describe.**
When scrolling down to read more issues, the User Profile photos are automatically hidden and can no longer be seen unless we scroll up to the top. Having the user profiles always in sight is important to know who creates which PR or issue. Below are two screenshots of full-screen application and the card views containing user profile photos are hidden.

<img width="1512" alt="image" src="https://user-images.githubusercontent.com/25452497/216889816-b9f54352-1ff1-4571-9b46-d5d7a12dc6ae.png">

<img width="1512" alt="image" src="https://user-images.githubusercontent.com/25452497/216889850-7fddbd60-b167-43be-8d68-70d5914bee16.png">

**Describe the solution you'd like**
Having the user profile mat-card always in sight while scrolling down while hiding other less important information like filter.

**Additional context**

chia-yh:
I think this solution would be a good improvement to the UI, however, I'd like to propose an alternative where the PR/issues are contained in scrollable containers for each author. This allows the filter bar to remain in sight, allowing users to be able to change the filter settings easily, while also being able to see the author of each PR/issue, which solves the problem raised. Having the PR/issues be in individual scrollable containers for each author would also mean that scrolling through the PR/issues of a single author won't potentially hide a PR/issue of another author that the user may be looking at.

To illustrate what I mean, I managed to tweak `card-view` component slightly to get the following result:
![image](https://github.com/CATcher-org/WATcher/assets/105497890/da2c6ffe-dc68-4d82-a182-d70ccbf0c446)
As shown above, users would be able to individually look through the PR/issues of a particular author, and there would also be no need for users to scroll up if they would like to change the filter settings. However, I'm not sure if this alternative would be an improvement over the originally proposed solution.

damithc:
> I think this solution would be a good improvement to the UI, however, I'd like to propose an alternative where the PR/issues are contained in scrollable containers for each author. This allows the filter bar to remain in sight, allowing users to be able to change the filter settings easily, while also being able to see the author of each PR/issue, which solves the problem raised. Having the PR/issues be in individual scrollable containers for each author would also mean that scrolling through the PR/issues of a single author won't potentially hide a PR/issue of another author that the user may be looking at.
> 
[...]

PR #148, opened 2023-06-25
Enclose PR/issues in scrollable container

chia-yh:
### Summary:
Possible alternative solution to the problem raised in #74, where the filter bar is also kept in sight, and users are able to scroll through the PR/issues of each author individually.

### Changes Made:
- Modify `card-view` component to enclose PR/issues of each author in individual scrollable containers

**Chrome**:
scroll_shadows_chrome

**Firefox**:
scroll_shadows_firefox

### Commit Message:

```
Enclose PR/issues in scrollable container

User profiles are automatically hidden when scrolling down to read more 
issues, and can no longer be seen without scrolling back up. 

Being able to see user profiles at all times helps with knowing the
author of each PR/issue.

Let's change it so that the user profile mat-card is kept in sight while
scrolling through PR/issues.
```


src/app/issues-viewer/card-view/card-view.component.html
```diff
@@ -18,12 +18,14 @@
     </mat-card>
   </div>
 
-  <div *ngFor="let issue of this.issues$ | async; index as i">
-    <app-issue-pr-card [issue]="issue" [dropdownFilter]="issues.dropdownFilter"></app-issue-pr-card>
-  </div>
+  <mat-card-content style="position: relative;">
```
gycgabriel, line 21: mat-card should enclose mat-card-content

src/app/issues-viewer/card-view/card-view.component.html
```diff
@@ -18,12 +18,14 @@
...
+    <div *ngFor="let issue of this.issues$ | async; index as i" class="scrollable-container" style="position: relative; z-index: -1;">
```
gycgabriel, line 22: Is this class used anywhere?

gycgabriel (changes requested):
Good work, thanks for the investigation to scroll shadows. 

Small nitpicks, and let's try to move styles into the css file as much as possible.

src/app/issues-viewer/card-view/card-view.component.html
chia-yh: I think I left it in there while I was experimenting, I'll remove it, sorry about that

src/app/issues-viewer/card-view/card-view.component.html
gycgabriel: Since we are not using mat-card for this, I think div is a better element to use

src/app/issues-viewer/card-view/card-view.component.html
chia-yh: I think I left this here while experimenting, I'll rename it to something more appropriate and move the styles to the css

src/app/issues-viewer/card-view/card-view.component.html
chia-yh: You're right, I'll work on tidying it up, thanks!

gycgabriel:
Even better if we can preserve the existing shadow or have some minimum black shadow:


image
image

chia-yh:
> Even better if we can preserve the existing shadow or have some minimum black shadow:
> 
> image image

Didn't notice the changes removed the shadows, thanks! I made some changes and I think I managed to restore the existing shadow by adjusting the `z-index` of the `column-header`:
image

gycgabriel (approved):
LGTM, thanks for the changes

PR #150, opened 2023-06-28
Add reset labels feature

### LINKED ISSUE #112: Small icon button to reset Label filter
Screenshot 2023-03-28 at 4 21 55 PM

Add a small icon to click to reset all hidden and selected labels. Easy reset.

**Additional context**
After filtering by many labels, or hiding labels, users may want a quick and easy way to reset the filter.
Icon can be placed on the right of the label search bar

joyngjr:
### Summary:
Fixes #112 

### Changes Made:

- Adds a remove all button to drop down filter list
- Calls a new method `removeAllSelection()`

Previously:
image

Currently:
image

### Commit Message:

```
Adds a small button to reset existing filters on issues

Previously, users have to unselect each issue tag individually.

Let's add a small button to reset existing filters.
```


src/app/shared/filter-bar/label-filter-bar/label-filter-bar.component.ts
```diff
@@ -93,4 +93,10 @@ export class LabelFilterBarComponent implements OnInit, OnDestroy {
   filter(filter: string, target: string): boolean {
     return !target.toLowerCase().includes(filter.toLowerCase());
   }
+
+  removeAllSelection(): void {
+    this.selectedLabels.next([]);
+    this.selectedLabelNames = [];
+    this.cdref.detectChanges();
+  }
```
cheehongw, line 101:
```suggestion
  @ViewChild(MatSelectionList) matSelectionList;   //add after L19
  
  removeAllSelection(): void {
    this.matSelectionList.deselectAll();
  }
```

cheehongw (changes requested):
`ChangeDetectorRef` appears to suppress this warning but it seems like there are still unintended effects. 

image

 
 
 After selecting 4 elements, clicking the remove all button clears the selection, but it still says 3 selected. (see video)

https://github.com/CATcher-org/WATcher/assets/72195240/2f257d3c-fedc-41e5-b0c2-6da1662ccafc

src/app/shared/filter-bar/label-filter-bar/label-filter-bar.component.ts
cheehongw: after spending a few hours investigating this, I thought it might be interesting to share what caused this `ExpressionChanged...Error`.

Angular throws an `ExpressionChangedAfterItHasBeenCheckedError` when an expression value has been changed after change detection has completed. (see: https://angular.io/errors/NG0100)

To begin you must first understand the stages in the component lifecycle (which includes change detection) - https://angular.io/guide/lifecycle-hooks#prerequisites

In this case, this is probably caused by a child component changing a parent's property. The parent would be `LabelFilterBarComponent`, the child would be `MatSelectionList`. The change detection is initiated by `this.selectedLabelNames = []`. The child then changes this property again through the two-way binding of `ngModel`:

```
<mat-selection-list [(ngModel)]="selectedLabelNames (selectionChange)="selectedLabels.next(selectedLabelNames)">
```

joyngjr:
I've made the changes mentioned!

src/app/shared/filter-bar/label-filter-bar/label-filter-bar.component.ts
```diff
@@ -1,5 +1,5 @@
-import { Component, Input, OnDestroy, OnInit } from '@angular/core';
-import { MatListOption } from '@angular/material/list';
+import { ChangeDetectorRef, Component, Input, OnDestroy, OnInit, ViewChild } from '@angular/core';
```
cheehongw, line 1: nit: since `ChangeDetectorRef` is not used, we should not include it in imports

src/app/shared/filter-bar/label-filter-bar/label-filter-bar.component.ts
```diff
@@ -25,7 +26,7 @@ export class LabelFilterBarComponent implements OnInit, OnDestroy {
 
   labelSubscription: Subscription;
 
-  constructor(private labelService: LabelService, private logger: LoggingService) {}
+  constructor(private labelService: LabelService, private logger: LoggingService, private cdref: ChangeDetectorRef) {}
```
cheehongw, line 29: nit: `cdref` not used

cheehongw (changes requested):
just a few more changes.

Other than that, LGTM

src/app/shared/filter-bar/label-filter-bar/label-filter-bar.component.ts
gycgabriel: I think it's better not to use cdref, and in this case we don't need it

src/app/shared/filter-bar/label-filter-bar/label-filter-bar.component.ts
```diff
@@ -93,4 +94,8 @@ export class LabelFilterBarComponent implements OnInit, OnDestroy {
   filter(filter: string, target: string): boolean {
     return !target.toLowerCase().includes(filter.toLowerCase());
   }
+
+  removeAllSelection(): void {
+    this.matSelectionList.deselectAll();
```
gycgabriel, line 99: There's a problem with this actually, it doesn't update the label filter on the issues, i.e. selectionChange doesn't run

After updating the view, call what's inside selectionChange on mat-selection-list, to run the behaviour of manually clicking the checkbox

gycgabriel: Don't forget to abstract that into another method so that there is no duplicated code

gycgabriel (changes requested):

joyngjr:
Apologies for the late changes, it should be working now.

gycgabriel (approved):
LGTM

cheehongw (approved):
LGTM!

PR #151, opened 2023-06-30
Show loading spinner on switch repository

chia-yh:
### Summary:
Fixes #35 
WATcher freezes before changing to new repository issues. After some investigation, I believe this delay is due to the checking of the validity of the repository using `isRepositoryPresent` in `changeRepositoryInPhaseIfValid` below:
```
    this.githubService.isRepositoryPresent(repo.owner, repo.name).subscribe((isValidRepository) => {
      if (!isValidRepository) {
        this.phaseService.isChangingRepo.next(false);
        throw new Error('Invalid repository name. Please check your organisation and repository name.');
      }
```
As suggested in the title of the issue, the fix implemented is to show a loading spinner while WATcher is changing repositories. This is indicated by a `BehaviorSubject` under `PhaseService`. `.next()` is used to indicate whether WATcher is changing repositories by updating `BehaviorSubject` before checking the validity of the repository, as well as after the repository is determined to be invalid or after the repository has been changed.

### Changes Made:

- Add `isChangingRepo` to `PhaseService` to indicate whether WATcher is changing repositories
- Modify `issue-viewer.component.html` to show a loading spinner depending on the value of `isChangingRepo`
- Modify `changeRepositoryInPhaseIfValid` in `header.component.ts` to update `isChangingRepo` accordingly

loading_spinner

### Commit Message:

```
Show loading spinner on switch repository

WATcher freezes before changing to new repository issues.

Let's show a loading spinner upon switching to a new repository.
```

PR #162, opened 2023-07-05
Extract method from header.component into phase.service.ts

chia-yh:
### Summary:

Addresses https://github.com/CATcher-org/WATcher/pull/151#discussion_r1249395736
The nature of `changeRepositoryInPhaseIfValid()` makes it more suited to be in `phase.service` instead of `header.component`.

### Changes Made:

- Extract the relevant logic from `HeaderComponent::changeRepositoryInPhaseIfValid()` into a new method `PhaseService::changeRepositoryIfValid()`, and rename the original method to `HeaderComponent::changeRepositoryIfValid()`
- Remove `switchRepo()` and move method calls into `PhaseService::changeRepositoryIfValid()` and `HeaderComponent::changeRepositoryIfValid()` where appropriate

### Commit Message:

```
Extract method from header.component

changeRepositoryInPhaseIfValid() is more suited to phase.service than
header.component.

Let's extract the method into phase.service where appropriate.
```

### DIFF
--- a/src/app/core/services/phase.service.ts
+++ b/src/app/core/services/phase.service.ts
@@ -76,20 +76,23 @@ export class PhaseService {
 
   /**
    * Changes current respository to a new repository.
-   * If on Issue Dashboard, add previously visited repositories to otherRepos.
    * @param repo New current repository
    */
-  changeCurrentRepository(repo: Repo): void {
+  private changeCurrentRepository(repo: Repo): void {
     this.logger.info(`PhaseService: Changing current repository to '${repo}'`);
 
     if (this.currentPhase === Phase.issuesViewer) {
       /** Adds past repositories to phase */
-      this.otherRepos.push(this.currentRepo); // TODO feature: can be used to provide repo suggestions
+      this.otherRepos.push(this.currentRepo);
     }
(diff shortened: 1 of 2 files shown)


gycgabriel (approved):
LGTM

src/app/core/services/phase.service.ts
```diff
@@ -90,6 +90,19 @@ export class PhaseService {
     this.repoChanged$.next(repo);
   }
 
+  async changeRepositoryIfValid(repo: Repo) {
```
gycgabriel, line 93: Let's add comments similar to the style in phase.service.ts

gycgabriel: Also make `changeCurrentRepository` private and remove comments to 
`If on Issue Dashboard, add previously visited repositories to otherRepos`
and 'provide repo suggestions' because the the implementation has been changed. 

This is so that it is clear to developers to use `changeRepositoryIfValid` to call `changeCurrentRepository`

gycgabriel (changes requested):
Some minor changes

gycgabriel (approved):
LGTM

Issue #117, opened 2023-04-03
Repeated calls to get Issues 


vigneshsankariyer1234567890:
**Describe the bug**
There are many repeated calls to get Issues from GitHub, as seen in the screenshot. This made sense for CATcher, as it was an application which required real-time data. However, WATcher does not need to refresh the issues every 5 seconds, and users can afford to just view non-updated data. 

This makes sense as we want to reduce the number of calls. 

<img width="656" alt="image" src="https://user-images.githubusercontent.com/68676914/229417597-6d4f1fba-c497-43d1-8f40-e35f7049d879.png">

**To Reproduce**
Login to WATcher and open network. Issues are polled every 5 seconds using `IssueService::startPollIssues`.

**Desktop (please complete the following information):**

- OS: MacOS Monterey
- Browser Chrome

seetohjinwei:
Hi! For this, do we want to just disable the issue polling entirely or reduce the latency to something much longer (maybe 5 minutes)?

Eclipse-Dominator:
This could be a possible solution, but by the nature of WATcher, it might be better to replace the polling with a manual refresh. (perhaps the a sync button).

PR #165, opened 2023-07-07
Update labels without refresh

### LINKED ISSUE #13: Issue Dashboard: Update labels without needing to refresh
**Current**
Fetch labels only OnInit of Label Chip Bar.

**Desired**
Fetch labels on github event change just as issue tables is updated and on Sync button click

chia-yh:
### Summary:

Attempt to resolve #13 

Labels are only fetched on initialization, which requires users to refresh the page in order for labels to update. Having labels update without requiring users to refresh would be ideal. The changes in this PR make it so that labels are polled and updated periodically, and also causes the Sync button to also update labels.

Currently polls label data at intervals of 5 seconds (the same `POLL_INTERVAL` as issues), however, as updates to labels are unlikely to occur as frequently as updates to issues, increasing the interval is probably a good idea to avoid excessive requests, as users can still use the Sync button if necessary.

#### Type of change:

- ✨ New Feature/ Enhancement

### Changes Made:

-`label.service.ts`: added periodic polling and updating of label data
-`label-filter-bar.component.ts`: use updated `label.service.ts` to support updating of label data, moved `simplifiedLabel` into `label.service.ts`
-`header-component.ts`: changed `reload()` to also fetch label data
-Handle setup/cleanup of `LabelService` where appropriate

### Screenshots:

Periodic polling:
update_labels_poll

Sync button (_polling interval adjusted to 30 seconds for demonstration purposes_):
update_labels_button

### Proposed Commit Message:

```
Update labels without needing to refresh

Labels are only fetched on initialization, requiring users to refresh
in order for labels to update.

Let's change it so labels are updated periodically, and when the Sync
button is clicked.
```

<details><summary>
<h3>Checklist:</h3>
</summary>

</details>


gycgabriel (dismissed):
LGTM

gycgabriel:
@chia-yh Thanks for your PR. 

@CATcher-org/2223s2 @damithc Any thoughts on what is a good poll interval for fetching labels, or should we restrict it to a manual Sync button to minimize the number of github API requests?

src/app/core/services/label.service.ts
```diff
@@ -13,6 +13,11 @@ const COLOR_DARKNESS_THRESHOLD = 0.184;
 const COLOR_BLACK = '000000'; // Dark color for text with light background
 const COLOR_WHITE = 'ffffff'; // Light color for text with dark background
 
+export type simplifiedLabel = {
+  name: string;
+  color: string;
+};
+
```
gycgabriel, line 20: For another issue/PR: This would be good to resolve why we need Label vs SimplifiedLabel, it seems redundant to have both

src/app/core/services/label.service.ts
```diff
@@ -22,17 +27,59 @@ const COLOR_WHITE = 'ffffff'; // Light color for text with dark background
...
   labels: Label[];
+  simplifiedLabels: simplifiedLabel[];
+
+  private labelsPollSubscription: Subscription;
+  private labelsSubject = new BehaviorSubject<simplifiedLabel[]>([]);
 
   constructor(private githubService: GithubService) {}
 
+  startPollLabels() {
+    if (this.labelsPollSubscription === undefined) {
```
Eclipse-Dominator, line 41: Might be clearer to users if we convert this to a guard. i.e.
```suggestion
    if (this.labelsPollSubscription) {
        return;
    }
```

src/app/core/services/label.service.ts
```diff
@@ -22,17 +27,59 @@ const COLOR_WHITE = 'ffffff'; // Light color for text with dark background
...
     return this.githubService.fetchAllLabels().pipe(
       map((response) => {
         this.labels = this.parseLabelData(response);
+        this.simplifiedLabels = this.labels.map((label) => {
+          return {
+            name: label.getFormattedName(),
+            color: label.color
+          };
+        });
+        this.labelsSubject.next(this.simplifiedLabels);
```
Eclipse-Dominator, line 82: I think at this point it is probably more efficient to just store getFormattedName()'s result as a property of Label instead of cloning the entire label. 

We can still keep simplifiedLabel but make Label implement the type simplified label so this conversion will no longer be needed.

src/app/core/services/label.service.ts
```diff
@@ -13,6 +13,11 @@ const COLOR_DARKNESS_THRESHOLD = 0.184;
 const COLOR_BLACK = '000000'; // Dark color for text with light background
 const COLOR_WHITE = 'ffffff'; // Light color for text with dark background
 
+export type simplifiedLabel = {
+  name: string;
+  color: string;
+};
```
Eclipse-Dominator, line 19: It might be more appropriate to move this to label.model instead of in label.service.

Eclipse-Dominator (commented):
Some initial review

Eclipse-Dominator (changes requested):
See above

src/app/core/services/label.service.ts
chia-yh: Not sure if I implemented the change requested here in the manner that was intended, but I made an attempt to do so, please let me know if I should make further changes

src/app/core/models/label.model.ts
```diff
@@ -4,26 +4,27 @@
 export class Label {
```
Eclipse-Dominator, line 4: I meant by 
```suggestion
export class Label implements SimplifiedLabel {
```

This way Label is also a simplified label. So in service there is no need to convert the label array into simplified label array as label array is already a simplified label array.

Although we do need to rename the references to SimplifiedLabel from name to formattedName.

gycgabriel: Let's rename SimplifiedLabel -> SimpleLabel

So it makes sense for Label to implement SimpleLabel as a more complex label

chia-yh: I see, that's a much better way of implementing SimplifiedLabel. I've made the changes as suggested in the most recent commit, and also renamed SimplifiedLabel -> SimpleLabel as mentioned.

I've also made some changes to accommodate #183 , as this PR modifies parts of the original `label-filter-bar.component.ts` used in that PR.

chia-yh:
> @chia-yh Thanks for your PR.
> 
> @CATcher-org/2223s2 @damithc Any thoughts on what is a good poll interval for fetching labels, or should we restrict it to a manual Sync button to minimize the number of github API requests?

I was looking through some of the other issues, and came across #117 , which brings up a good point, that WATcher users can afford to view non-updated data. As this PR already has the sync button also update the labels, which aligns with the suggestion in https://github.com/CATcher-org/WATcher/issues/117#issuecomment-1616758747, should I change it so that the periodic polling of labels is removed, and the fetching of labels is restricted to just the sync button as mentioned in an earlier comment? 

While I think having periodic polling would be good to have, in my opinion, setting it to a long enough interval to prevent too many calls to fetch data would cause it to provide little benefit to the user, as they are unlikely to run into a situation where the refresh would be timely enough to update the data as they need it, and they could just use the sync button to refresh data.

Also, since the issue concerning periodic label polling is similar to the issue brought up in #117 , should the same changes decided on here be applied to the issue polling? I think the sync button also fetches issues at the moment, so it would just be a matter of removing the polling at 5 second intervals, though the change should definitely be done in a separate PR addressing #117 directly.

Eclipse-Dominator:
Perhaps we can make the polling an optional toggle somewhere on the page. 
This way users (who somehow want to keep the page open for a long time) can still have access to updated data. Handling it should be done in a seperate PR.

Eclipse-Dominator (approved):
LGTM!

gycgabriel (approved):
LGTM

PR #179, opened 2023-07-17
Immediately display page data that has loaded

### LINKED ISSUE #83: Asynchronous loading of page data
Currently, the content of the page is loaded synchronously from the repositories via multiple requests. This does not present much of an issue most of the time but it can take a rather long time if there are many pages of PRs and issues to be loaded in especially when there is a lot of PRs and Issues in a repository. 

Propose the use of asynchronous loading of data to immediately load in the page data once the response is received. We can use a variable to keep track of doneness of loading of data. as well as to show the loading symbol to indicate data is still being loaded in. This enables the page to immediately load in the data once received and it should make the page feel smoother and more responsive.

seetohjinwei:
### Summary:

Fixes #83

#### Type of change:

- ✨ New Feature/ Enhancement
- 🎨 Code Refactoring

### Changes Made:

- Updated `withPagination` to continually emit the currently accumulated results as it new data is received.
- Updated relevant portions to support this change.

### Screenshots:

https://github.com/CATcher-org/WATcher/assets/47915290/e4b086ec-5550-4bf3-8fd0-a0dffcb74363

### Proposed Commit Message:

```
Immediately display page data that has loaded

Previously, the page would show loading spinners, with no content,
until all the data has been loaded.

Let's
* Display the page data once it is loaded.
* Display loading spinners below the content, until all data has been
  loaded.
* Change `Sync` to a spinner while the content is being loaded.
```

<details><summary>
<h3>Checklist:</h3>
</summary>

</details>

Sources referenced:

- https://stackoverflow.com/a/68229248 (ended up using the approach mentioned here)
- https://www.typescriptlang.org/docs/handbook/release-notes/typescript-3-6.html
- https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/AsyncGenerator
- https://www.apollographql.com/docs/react/data/refetching/
- https://xgrommx.github.io/rx-book/content/getting_started_with_rxjs/creating_and_querying_observable_sequences/generators_and_observable_sequences.html

### DIFF
--- a/src/app/core/services/github.service.ts
+++ b/src/app/core/services/github.service.ts
@@ -483,7 +483,7 @@ export class GithubService {
     pluckEdges: (results: ApolloQueryResult<T>) => Array<any>,
     Model: new (data) => M
   ): Observable<Array<M>> {
-    return this.withPagination<T>(pluckEdges, query, variables).pipe(
+    return this.withPagination<T>(pluckEdges, query, variables, false).pipe(
       map((results: ApolloQueryResult<T>[]) => {
         const issues = results.reduce((accumulated, current) => accumulated.concat(pluckEdges(current)), []);
         return issues.map((issue) => new Model(issue.node));


src/app/core/services/github.service.ts
```diff
@@ -495,32 +495,55 @@ export class GithubService {
...
+
+        // Use a chain of await to ensure that all recursive queries are completed before `complete` is called.
+        await queryWith(nextCursor);
       });
+    }
+
+    const initialCursor = null;
+    const observer = async () => {
+      await queryWith(initialCursor);
+      behaviorSubject.next(accumulatedResults);
```
Eclipse-Dominator, line 541:
```suggestion
```

Not neccessary, before `queryWith` ends, the last `accumulatedResults` is already updated

src/app/core/services/github.service.ts
```diff
@@ -495,32 +495,55 @@ export class GithubService {
...
+      behaviorSubject.complete();
     };
+
+    observer();
```
Eclipse-Dominator, line 545: It's possible to move `behaviorSubject.complete()` into `queryWith()`'s return condition.

This simplifies this async from call to 
```suggestion
    queryWith(null)
```

Eclipse-Dominator (changes requested):
Features work as expected, I have included some minor observations.

Some additional observation
> - Display loading spinners below the content, until all data has been
  loaded.

for the paginators, when the current page is full, the loading spinner disappears which may misinform users if there are still issues to be loaded. 

Perhaps also include some additional way to indicate that some data is still being loaded like via the sync button to inform the users that data is still being fetched from the servers.

src/app/core/services/github.service.ts
seetohjinwei: Without the final `behaviorSubject.next(accumulatedResults)`, it seems like the final batch of PRs wouldn't actually be sent. However, I'm not entirely sure why this is the case :/

Reproducible by removing this line, going to `catcher-org/catcher`, and searching for PR `1203` (one of the newer PRs)

I tried making the following change, together with the other comment, but it's still facing the same issue

```
M src/app/core/services/github.service.ts
@@ -527,6 +527,7 @@ export class GithubService {
         behaviorSubject.next(accumulatedResults);
         if (edges.length < maxResultsCount || !nextCursor) {
           // No more queries to perform.
+          behaviorSubject.complete();
           return;
         }
 
@@ -535,14 +536,15 @@ export class GithubService {
       });
     }
 
-    const initialCursor = null;
-    const observer = async () => {
-      await queryWith(initialCursor);
-      behaviorSubject.next(accumulatedResults);
-      behaviorSubject.complete();
-    };
-
-    observer();
+    queryWith(null);
 
     return observable;
   }
```

src/app/core/services/github.service.ts
seetohjinwei: Related to the other comment, but if we need to keep the `next` and `complete` here, maybe we could use a `.then` chain instead?

```js
const initialCursor = null;
queryWith(initialCursor).then(() => {
  behaviorSubject.next(accumulatedResults);
  behaviorSubject.complete();
});
```

seetohjinwei:
I've made the change in 3305adb to swap out the `Sync` button for a `Loading` message!

However, I noticed that `isLoading` (in `src/app/core/services/issue.service.ts`) is re-assigned, which causes `isLoading$` (that was added in this commit) to not be triggered on a repo change. So, I made some changes in that file too. I think the other components that are using this observable shouldn't be affected by this change, but I'm not entirely sure about this!

https://github.com/CATcher-org/WATcher/assets/47915290/3d487474-82b8-4229-aa27-1d538f8148ee

Eclipse-Dominator:
Alright, I did some investigation regarding why the additional `behaviorSubject.next()` is needed.

The cause of the issue is in `fetchIssuesGraphql()`
```typescript
  fetchIssuesGraphql(issuesFilter: RestGithubIssueFilter): Observable<Array<GithubIssue>> {
    // ...
    const issueObs = this.toFetchIssues(issuesFilter).pipe(
      // observable that fetches Issues
    );
    const prObs = this.toFetchIssues(issuesFilter).pipe(
      // observable that fetches PRs
    );

    // Concatenate both streams together.
    return zip(issueObs, prObs).pipe(map((x) => x[0].concat(x[1])));  // <----- this line
  }
```
[zip](https://www.learnrxjs.io/learn-rxjs/operators/combination/zip) will only emit a value when all input observables have emitted a value. 

This means when both `issueObs` and `prObs` ends, one of the Observable may have emitted less value than the other.
This causes the issue of missing data.

However, just emitting the observable an additional time may not actually load all the data. As in a hypothetical repo where there are a lot more PRs than issues or vice versa, it is possible for some PRs/Issues to be left out and not be saved. 

Previously, this is okay because `withPagination` only returns a single result so the pairing works

Fix
---
The proper way to concatenate the 2 observable streams is to replace `zip` with `merge`. 

However, doing so will lead to the current page deleting itself when new data is being loaded.

https://github.com/CATcher-org/WATcher/assets/4567895/a25baf55-3be8-4d53-8a61-15b73b85a013

The resulting behavior is caused by:

```typescript
  private initializeData(): Observable<Issue[]> {
    let issuesAPICallsByFilter: Observable<Array<GithubIssue>>;
    // ...    
    return issuesAPICallsByFilter.pipe(
      map((issuesByFilter: []) => {
        const fetchedIssueIds: Array<Number> = [];

        // Take each issue and put it in next in issues$
        for (const issue of issuesByFilter) {
          fetchedIssueIds.push(this.createIssueModel(issue).id);
          this.createAndSaveIssueModel(issue);
        }

        const outdatedIssueIds: Array<Number> = this.getOutdatedIssueIds(fetchedIssueIds);  // <--- these lines
        this.deleteIssuesFromLocalStore(outdatedIssueIds);   // <-- these lines

        if (this.issues === undefined) {
          return [];
        }
        return Object.values(this.issues);
      })
    );
  }
```

Previously, both issues and prs are combined and received at this point of the code, `getOutdatedIssuesIds` will filter away Issues and PRs that no longer exist in the repo. (i.e. the relevant issue/pr is being deleted/transferred from the repo) done by the actions below.
image

However, the data is now being loaded in by types (PR/Issue) since the code is now swapped to using `merge`.
Loading the PR list will cause all issues stored in `issue.service` to be deleted and loading of the Issue list will cause the PRs to be deleted thus creating this 'toggling' effect.

To resolve this, we can make `fetchedIssueIds` a global variable in `initializeData()` so that it will always keep track of all current fetched issues in the repo

```typescript
  private initializeData(): Observable<Issue[]> {
    let issuesAPICallsByFilter: Observable<Array<GithubIssue>>;
    const fetchedIssueIds: Array<Number> = []; // <--
    // ...    
    return issuesAPICallsByFilter.pipe(
      map((issuesByFilter: []) => {
        // const fetchedIssueIds: Array<Number> = [];
```

---
side note, I think the type for `fetchedIssueIds` should be `number[]`? didn't really look into it.

src/app/core/services/github.service.ts
Eclipse-Dominator: see [comment](https://github.com/CATcher-org/WATcher/pull/179#issuecomment-1654671675)

src/app/core/services/github.service.ts
```diff
@@ -495,32 +495,55 @@ export class GithubService {
...
+    pluckEdges: (results: ApolloQueryResult<T>) => Array<any>,
+    query: DocumentNode,
+    variables: { [key: string]: any } = {}
+  ): Observable<ApolloQueryResult<T>[]> {
+    const maxResultsCount = 100;
+    const apollo = this.apollo;
+
+    let accumulatedResults: ApolloQueryResult<T>[] = [];
+    const behaviorSubject: BehaviorSubject<ApolloQueryResult<T>[]> = new BehaviorSubject(accumulatedResults);
+    const observable = behaviorSubject.asObservable();
```
Eclipse-Dominator, line 516:
```suggestion
```
This variable is not used throughout the function. I think we can just do 
```typescript
return behaviorSubject.asObservable();
```
to make te method less cluttered.

Eclipse-Dominator (changes requested):
LGTM except for the issue mentioned in the comment above.

I think the syncing looks good! 

Personally, I think showing the 'loading' text is not necessary and just having the spinner icon will be enough. 

Let's also update this branch with the main branch as well just to be safe.

seetohjinwei:
Thank you! I did attempt to use `merge` at some point, but didn't realise that `getOutdatedIssueIds` was the one causing the "older" entries to get cleared off! I'm still not very familiar with Angular observables, so I thought I misunderstood the docs on `zip` and `merge`. Really appreciate the detailed explanation!

> However, it does seem to be more appropriate if we do the check here instead which might take more work but so far I havn't really noticed anything majorly wrong with the simple solution above during testing.

I couldn't really find a way to get the value emitted in `finalize` or `tap`'s `completion` callback. I think it's because the `finalize` could potentially be triggered from an error, so there might not be a value emitted. I think the current implementation shouldn't be too expensive to run, as the number of issues is typically fairly small

> side note, I think the type for fetchedIssueIds should be number[]? didn't really look into it.

I did some reading on the docs, seems like

- `Array<number>` and `number[]` are effectively the same
  - https://www.typescriptlang.org/docs/handbook/2/generics.html#working-with-generic-type-variables
- the use of `Number` as a type is discouraged
  - https://www.typescriptlang.org/docs/handbook/2/everyday-types.html#the-primitives-string-number-and-boolean

I've changed the occurrences of `Array<Number>` to` number[]`.

---

Changes made:

* Synchronised the branches
* Remove `Loading` text
* Update `initializeData` to properly handle the pagination (using `merge` and moving `fetchedIssueIds` outside the arrow function as suggested).
* Change `Number` -> `number`

Eclipse-Dominator (approved):
LGTM!
Let's wait for at least one more people to review this before merging.

Eclipse-Dominator (changes requested):
Sorry for another review, I noticed some areas where there could be some improvements in efficiency.

src/app/core/services/github.service.ts
```diff
@@ -495,32 +495,49 @@ export class GithubService {
...
+    async function queryWith(cursor: string): Promise<void> {
+      const graphqlQuery = apollo.watchQuery<T>({ query, variables: { ...variables, cursor } });
+
+      await graphqlQuery.refetch().then(async (results: ApolloQueryResult<T>) => {
         const intermediate = Array.isArray(results) ? results : [results];
         const edges = pluckEdges(results);
         const nextCursor = edges.length === 0 ? null : edges[edges.length - 1].cursor;
 
+        accumulatedResults = accumulatedResults.concat(intermediate);
+        behaviorSubject.next(accumulatedResults);
```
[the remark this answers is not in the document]
seetohjinwei: I see! I've made the changes

src/app/core/services/github.service.ts
```diff
@@ -483,8 +483,8 @@ export class GithubService {
     pluckEdges: (results: ApolloQueryResult<T>) => Array<any>,
     Model: new (data) => M
   ): Observable<Array<M>> {
-    return from(this.withPagination<T>(pluckEdges)(query, variables)).pipe(
-      map((results: Array<ApolloQueryResult<T>>) => {
+    return this.withPagination<T>(pluckEdges, query, variables).pipe(
```
Eclipse-Dominator, line 486:
```suggestion
    return this.withPagination<T>(pluckEdges, query, variables, false).pipe(
```

Eclipse-Dominator (changes requested):
LGTM besides the listed change.

src/app/core/services/github.service.ts
seetohjinwei: I've added this, I initially misunderstood your previous comment, sorry!

Eclipse-Dominator (approved):
LGTM!

PR #182, opened 2023-07-20
Filter for 'Merged'

### LINKED ISSUE #167: No way to filter by merged PRs
**Describe the bug**
Status - Closed filters closed PRs, doesn't show merged PRs. 

**Expected behavior**
A way to filter by merged PRs - maybe an additional status "Merged"?

joyngjr:
### Summary:

Fixes #167 

Currently, PRs cannot be filtered by its 'merged' status even though merged PRs are displayed when no filters are set. 

However, as issues should only be filtered with status 'all/open/closed', 'all' is used as a default status when user tries to achieve an invalid combination (status: merged + type: all/PR -> status: merged + type: issue).

Option to select 'merged' status is removed when the current selected type is 'issue'.

#### Type of change:

- ✨ New Feature/ Enhancement
- 🐛 Bug Fix

### Changes Made:

- Adds an option to filter by 'merged' status for types 'all' and 'pull request'

### Screenshots:

https://github.com/CATcher-org/WATcher/assets/97519138/8ca45598-0fb6-4642-8580-803dcc6c67cd

### Proposed Commit Message:

```
Allow users to filter their PRs by 'Merged' status.

Previously, merged PRs are only shown when status is selected as 'All'.

Let's allow users to filter their PRs by 'Merged' status.
```

<details><summary>
<h3>Checklist:</h3>
</summary>

</details>


src/app/shared/filter-bar/filter-bar.component.ts
```diff
@@ -80,13 +80,29 @@ export class FilterBarComponent implements OnInit, AfterViewInit, OnDestroy {
...
 
+  /**
+   * Changes status to a valid, default value when an incompatible combination of type and status is encountered.
+   */
+  checkValidCombination() {
+    if (this.dropdownFilter.status === 'merged' && this.dropdownFilter.type === 'issue') {
+      this.dropdownFilter.status = 'all';
+    }
+  }
+
```
gycgabriel, line 91: Could we add some logic to change the dropdown filter type to Pull Request if Merged is selected? 

I don't think we need this logic? Because it seems impossible to select Merged when Issue is selected

gycgabriel (changes requested):
Small changes

src/app/shared/filter-bar/filter-bar.component.ts
joyngjr: I've changed it so that PR will be selected if merged is selected!

I think we will still need to check for the combination 'issue + merged' because users can select 'pr + merged' and then change the type from pr to issue. 

Another possibility could be to remove the 'issue' option when 'merged' is selected, but this would mean more coupling and I'm not sure if this would cause the behaviour to deviate too much from the original.

gycgabriel: Better to not hide the Issue option, since that filter is more important than filtering by status. Thanks this implementation is good

gycgabriel (approved):
LGTM

PR #187, opened 2023-07-29
Throw error if there are too many pages to fetch

### LINKED ISSUE #32: Timeout after certain duration of graphql query and show error message
For repositories larger than 200 issues or pull requests, WATcher may hang or crash from loading.

Timeout after certain duration of graphql query and show error message

See https://github.com/CATcher-org/WATcher/issues/22#issuecomment-1193159356

seetohjinwei:
### Summary:

Fixes #32

#### Type of change:

- ✨ New Feature/ Enhancement
- 🐛 Bug Fix

### Changes Made:

- Change `toFetchIssues` to throw an error when the number of pages in a repository is more than `100`.
- Made change to `issue.service.ts` because to propagate the error upwards instead of silently returning an empty observable.
- After making the previous change, I noticed that there was an error thrown on empty repositories, which had previously been silently ignored, so I've fixed it by doing a simple conditional check with `this.issues === undefined`. (I've made the same change on #179 too)

I noticed that the application crashed within `toFetchIssues` when trying to make the REST api calls, i.e. it didn't even reach the graphql API stage. The REST api calls were made concurrently, so trying to cancel the already-made requests on a timeout would be more complicated, and does not have much benefits over a pre-determined limit. So, I made the check on the number of pages instead of a timeout for the graphql API.

### Screenshots:

https://github.com/CATcher-org/WATcher/assets/47915290/8280482b-b5bc-41eb-8118-8330275b6e18

- Tested on large repository `angular/angular`
- Tested on `catcher-org/catcher`, `catcher-org/watcher` (small to medium sized repositories)
- Tested on empty repository `seetohjinwei/catcher-test`

### Proposed Commit Message:

```
Throw error if there are too many pages to fetch

Previously, if a repository with too many pages was opened, the
application would hang or crash. This would not be a pleasant
experience if a user were to accidentally open such a repository.

Let's
* Detect this scenario and handle it elegantly with an error message.
```

<details><summary>
<h3>Checklist:</h3>
</summary>

</details>

### DIFF
--- a/src/app/core/services/github.service.ts
+++ b/src/app/core/services/github.service.ts
@@ -148,6 +148,8 @@ export class GithubService {
    * @returns Observable<boolean> that returns true if there are pages that do not exist in the cache model.
    */
   private toFetchIssues(filter: RestGithubIssueFilter): Observable<boolean> {
+    const pageFetchLimit = 100;
+
     let responseInFirstPage: GithubResponse<GithubIssue[]>;
     return this.getIssuesAPICall(filter, 1).pipe(
       map((response: GithubResponse<GithubIssue[]>) => {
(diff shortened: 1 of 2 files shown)


gycgabriel (approved):
LGTM

PR #193, opened 2023-08-06
Improve efficiency of saving and deleting issue models

### LINKED ISSUE #188: Inefficient saving of Issues and PRs
**Describe the bug**

Methods `updateLocalStore()` and `deleteFromLocalStore()` in `issue.service` runs in O(n) times.
``` typescript
  /**
   * This function will update the issue's state of the application. This function needs to be called whenever a issue is added/updated.
   */
  updateLocalStore(issueToUpdate: Issue) {
    this.issues = {
      ...this.issues,
      [issueToUpdate.id]: issueToUpdate
    };
    this.issues$.next(Object.values(this.issues)); // <-- this takes up a lot of time
  }
```

seetohjinwei:
### Summary:

Fixes #188 

#### Type of change:

- ✨ New Feature/ Enhancement
- 🎨 Code Refactoring

### Changes Made:

- Remove duplicate usage of `createIssueModel`
  - in 80be52fbb61c0f17010729a2c10f40ffe893bf74
- Change `updateLocalStore` to take in an array of issues to avoid having to do `O(n)` loops for each issue
  - in 80be52fbb61c0f17010729a2c10f40ffe893bf74
- Change `deleteIssuesFromLocalStore` to directly update `this.issues` using `delete` keyword, remove currently unused `deleteFromLocalStore`
  - in 2531a1e6a2f97c0a7d85fac5824693cd60a84b39

### Screenshots:

N/A

### Proposed Commit Message:

```
Improve efficiency of saving and deleting issue models

Previously, saving and deleting issues took up to O(n^2).

Let's
* Remove unnecessary loops to achieve O(n).
```

<details><summary>
<h3>Checklist:</h3>
</summary>

</details>

### DIFF
--- a/src/app/core/services/issue.service.ts
+++ b/src/app/core/services/issue.service.ts
@@ -76,7 +76,7 @@ export class IssueService {
   getLatestIssue(id: number): Observable<Issue> {
     return this.githubService.fetchIssueGraphql(id).pipe(
       map((response: GithubIssue) => {
-        this.createAndSaveIssueModel(response);
+        this.createAndSaveIssueModels([response]);
         return this.issues[id];
       }),
       catchError((err) => {
@@ -96,13 +96,21 @@ export class IssueService {
 
   /**
    * This function will update the issue's state of the application. This function needs to be called whenever a issue is added/updated.
+   *
+   * @params issuesToUpdate - An array of issues to update the state of the application with.
+   * @params shouldEmit - Whether the updated issues should be emitted to issues$.
    */
-  updateLocalStore(issueToUpdate: Issue) {
-    this.issues = {
-      ...this.issues,
-      [issueToUpdate.id]: issueToUpdate
-    };
-    this.issues$.next(Object.values(this.issues));
+  updateLocalStore(issuesToUpdate: Issue[], shouldEmit: boolean = true) {
+    const newIssues = issuesToUpdate.reduce((obj, issue) => {
+      obj[issue.id] = issue;
+      return obj;
+    }, {});
+
+    this.issues = { ...this.issues, ...newIssues };
+
+    if (shouldEmit) {
+      this.issues$.next(Object.values(this.issues));
+    }
   }
 
   reset(resetSessionId: boolean) {
@@ -135,16 +143,15 @@ export class IssueService {
         return of([]);
     }
 
-    // const issuesAPICallsByFilter = filters.map(filter => this.githubService.fetchIssuesGraphql(filter));
     return forkJoin(issuesAPICallsByFilter).pipe(
-      map((issuesByFilter: [][]) => {
+      map((issuesByFilter: GithubIssue[][]) => {
         const fetchedIssueIds: Array<Number> = [];
 
         // Take each issue and put it in next in issues$
-        for (const issues of issuesByFilter) {
+        for (const githubIssues of issuesByFilter) {
+          const issues = this.createAndSaveIssueModels(githubIssues);
           for (const issue of issues) {
-            fetchedIssueIds.push(this.createIssueModel(issue).id);
-            this.createAndSaveIssueModel(issue);
+            fetchedIssueIds.push(issue.id);
           }
         }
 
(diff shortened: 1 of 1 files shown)


src/app/core/services/issue.service.ts
```diff
@@ -76,7 +76,7 @@ export class IssueService {
   getLatestIssue(id: number): Observable<Issue> {
     return this.githubService.fetchIssueGraphql(id).pipe(
       map((response: GithubIssue) => {
-        this.createAndSaveIssueModel(response);
+        this.createAndSaveIssueModels([response]);
         return this.issues[id];
       }),
       catchError((err) => {
```
vigneshsankariyer1234567890, line 82: The methods `getLatestIssue` and `getIssue` are not in use throughout the application. Let's take a look at unused methods in a separate issue and remove them from the codebase in phases.

src/app/core/services/issue.service.ts
```diff
@@ -85,24 +85,22 @@ export class IssueService {
...
    */
-  updateLocalStore(issueToUpdate: Issue) {
-    this.issues = {
-      ...this.issues,
-      [issueToUpdate.id]: issueToUpdate
-    };
-    this.issues$.next(Object.values(this.issues));
+  updateLocalStore(issuesToUpdate: Issue[], shouldEmit: boolean = true) {
+    const newIssues = { ...this.issues };
+    issuesToUpdate.forEach((issue) => {
```
vigneshsankariyer1234567890, line 96: This method is only in use within the class, and the method that is referencing it is also a private method (`createAndSaveIssueModels`). Let's make this a private method as well.

src/app/core/services/issue.service.ts
```diff
@@ -85,24 +85,22 @@ export class IssueService {
...
+   * @params issuesToUpdate - An array of issues to update the state of the application with.
+   * @params shouldEmit - Whether the updated issues should be emitted to issues$.
    */
-  updateLocalStore(issueToUpdate: Issue) {
-    this.issues = {
-      ...this.issues,
-      [issueToUpdate.id]: issueToUpdate
-    };
-    this.issues$.next(Object.values(this.issues));
+  updateLocalStore(issuesToUpdate: Issue[], shouldEmit: boolean = true) {
```
vigneshsankariyer1234567890, line 94: Since `shouldEmit` is true by default, and all methods which are calling `createAndSaveIssueModels` (`initializeData` and `getLatestIssue`) do not pass the `shouldEmit` parameter, it seems that we do not need the `shouldEmit` parameter in the first place. Besides, if there are any updates to the state, changes should be made known to the rest of the application; so let's look at removing the shouldEmit parameter for now until we find a use case for needing to selectively emit issues which are updated.

src/app/core/services/issue.service.ts
```diff
@@ -154,16 +151,29 @@ export class IssueService {
     );
   }
 
-  private createAndSaveIssueModel(githubIssue: GithubIssue): boolean {
-    const issue = this.createIssueModel(githubIssue);
-    this.updateLocalStore(issue);
-    return true;
+  private createAndSaveIssueModels(githubIssues: GithubIssue[], shouldEmit: boolean = true): Issue[] {
```
vigneshsankariyer1234567890, line 154: As mentioned in the comment regarding `updateLocalStore`, let's look at removing the shouldEmit flag.

src/app/core/services/issue.service.ts
```diff
@@ -154,16 +151,29 @@ export class IssueService {
...
 
-  private deleteIssuesFromLocalStore(ids: number[]): void {
-    ids.forEach((id: number) => {
-      this.getIssue(id).subscribe((issue) => this.deleteFromLocalStore(issue));
-    });
+  private deleteIssuesFromLocalStore(ids: number[], shouldEmit: boolean = true): void {
+    const withoutIssuesToRemove = { ...this.issues };
+    for (const id of ids) {
+      delete withoutIssuesToRemove[id];
+    }
```
vigneshsankariyer1234567890, line 170: Good idea to construct the object with all the issues and then remove issues from the object at once, rather than to construct an object with all issues and then remove a single issue for every issue to be removed.

vigneshsankariyer1234567890 (changes requested):
Thank you for your PR! Nicely done 🔥 

Changes are working well, so far LGTM 🚀 🚀  other than the small comments to resolve!

src/app/core/services/issue.service.ts
seetohjinwei: I think the original idea was that we can have some sort of function that has this kind of a calling behaviour:

```
shouldEmit = false
shouldEmit = false
...
shouldEmit = true  # at the end
```

I've removed it, I agree that it can be added back if this behaviour is needed in the future!

src/app/core/services/issue.service.ts
chunweii: Yep, there can be a separate PR that contributes to issue #72

vigneshsankariyer1234567890 (approved):
Approved :)

PR #207, opened 2023-09-18
Add Issues Dashboard access by URL 

### LINKED ISSUE #94: Add Issues Dashboard access by URL
Currently, users cannot access a specific repo's dashboard directly by URL. This means that Prof is not able to create a new URL for each team to distribute the link to each team's individual dashboard. Let's add the ability to access the relevant repository's dashboard by URL. 

Let's try to allow WATcher to grab the relevant repository's details using URL params. E.G. `/issuesViewer/{org-repo}`. Subsequently, we have 2 cases: either the user is logged in or not. We can store the relevant details in sessionStorage and extract it from sessionStorage once log-in done.

LJXSean:
### Summary:

Fixes #94 

#### Type of change:

- ✨ New Feature/ Enhancement

### Changes Made:

- Add `CacheRepoFromUrlService` that provides getter and setter for repoLocation stored in session storage
- Add `ParseUrlParamsGuard::canActivate` that uses `CacheRepoFromUrlService` to cache the 'org' and 'repo' params from URL into session storage
- Add path for '/issuesViewer/:org/:repo' to `app-routing` that uses `ParseUrlParamsGuard::canActivate`
- Edit `sessionSelectionComponent::autofillRepo` to fill and submit repoform if repoLocation is in session storage
- Subscription to `currentAuthState` in `HeaderComponent` now calls `openChangeRepoDialog` only if user is authenticated and repoLocation not in session storage
- `NOT_AUTHENTICATED_ERROR` now thrown for unauthenticated access on protected routes

### Screenshots:

With valid repo location
fix-94-success - Trim

With invalid repo location
fix-94-fail - Trim

### Proposed Commit Message:

```
Add path /issuesViewer/:org/:repo

Having functionality to get repo location from url parameters
allows easy distribution of links to a team's repository dashboard.

Let's store the repo location from the new path into session storage
and auto submit the session selection form.
```

<details><summary>
<h3>Checklist:</h3>
</summary>

</details>


LJXSean:
Some implementation details to discuss:

1. Session Storage persistance after login
- Possible issue if user logs out and tries to log in to select a different repository, but unable to do so due to session storage persistence that causes session selection form to autofill and submit
- Maybe reset session storage once form is filled?

2. Leftover code from CATcher
- eg. auth::getRepoFromUrlQueryParams and @input(): urlEncodedRepo
- No such repo query, should such code be removed?

3. Scope of "Login Required" error 
- Used to give user feedback on why redirect to login when accessing /issuesViewer/:org/:repo
- Currently implemented in authguard, introduces new behavior for displaying error for all unauthorised access
- Should this error be shown for all redirects or just that path

4. Path 'issuesViewer/:org/:repo' always redirects to login even if previously logged in before entering new link
- Follows current behavior where authentication state not persistant on page refresh
- eg. WATcher logs out on refresh, not sure if bug or desired behavior
- Should we make authentication state persistant?

5. Rename repoUrlCacheService
- confusing to have 2 caches for repoUrl that serves differnt purposes
- Should we as rename it as SuggestionService?

gycgabriel:
For point 4, this is an issue from CATcher -- https://github.com/CATcher-org/CATcher/discussions/709

For point 2 (and perhaps 5 as well), do consider creating another PR/Issue for that

src/app/app-routing.module.ts
```diff
@@ -3,10 +3,12 @@ import { RouterModule, Routes } from '@angular/router';
 import { ActivityDashboardModule } from './activity-dashboard/activity-dashboard.module';
 import { AuthModule } from './auth/auth.module';
 import { AuthGuard } from './core/guards/auth.guard';
+import { ParseUrlParamsGuard } from './core/guards/parse-url-params.guard';
 import { IssuesViewerModule } from './issues-viewer/issues-viewer.module';
 
 const routes: Routes = [
   { path: '', loadChildren: () => AuthModule },
+  { path: 'issuesViewer/:org/:repo', canActivate: [ParseUrlParamsGuard], children: [] },
```
vigneshsankariyer1234567890, line 11: Good use of guard function here, makes it pretty easy to read

src/app/shared/layout/header.component.ts
```diff
@@ -67,7 +69,7 @@ export class HeaderComponent implements OnInit {
       });
 
     this.auth.currentAuthState.subscribe((state) => {
-      if (auth.isAuthenticated()) {
+      if (auth.isAuthenticated() && !cacheRepoFromUrlService.hasRepoLocation()) {
```
vigneshsankariyer1234567890, line 72: Upon testing, I realised that changing repositories actually leads to having to authenticate again (as you mentioned in your comment). I think it might be good to persist session to sessionStorage in order for better UI/UX. However, I think this works pretty well by itself in the mean time, so let's add it as a separate issue and work on it after this is merged.

src/app/core/guards/parse-url-params.guard.ts
```diff
@@ -0,0 +1,22 @@
...
+  canActivate(route: ActivatedRouteSnapshot, state: RouterStateSnapshot) {
+    const org = route.params['org'];
+    const repo = route.params['repo'];
+    const repoLocation = `${org}/${repo}`;
+
+    this.cache.repoLocation = repoLocation;
+
+    this.router.navigateByUrl(Phase.issuesViewer);
+    return false;
+  }
```
vigneshsankariyer1234567890, line 21: Unusual to have both side effects (caching) and navigation although definitely wrong! Perhaps it might be better to add some comments or documentation to why this was done (as it took some time for me to understand the logic here behind intercepting the repo location and then rerouting to `issuesViewer`)

src/app/shared/layout/header.component.ts
```diff
@@ -67,7 +69,7 @@ export class HeaderComponent implements OnInit {
       });
 
     this.auth.currentAuthState.subscribe((state) => {
```
vigneshsankariyer1234567890, line 71: Since we don't check state, perhaps we can remove state here.
```suggestion
    this.auth.currentAuthState.subscribe(() => {
```

src/app/core/services/cache-repo-from-url.service.ts
```diff
@@ -0,0 +1,21 @@
+import { Injectable } from '@angular/core';
+
+@Injectable({
+  providedIn: 'root'
+})
+export class CacheRepoFromUrlService {
```
vigneshsankariyer1234567890, line 6: This has a very similar name to `RepoUrlCacheService`, maybe we could rename it to `RepoSessionStorageService`? This allows for lesser confusion esp within the SessionSelectionComponent and is a more direct approach, indicating it's about storing a repo within session storage.

src/app/auth/session-selection/session-selection.component.ts
```diff
@@ -117,6 +115,11 @@ export class SessionSelectionComponent implements OnInit {
   }
 
   private autofillRepo() {
-    this.repoForm.get('repo').setValue(this.urlEncodedRepo);
+    const repoLocation = this.cacheRepoFromUrlService.repoLocation || this.urlEncodedRepo;
+    this.repoForm.get('repo').setValue(repoLocation);
+
+    if (this.cacheRepoFromUrlService.hasRepoLocation()) {
+      this.setupSession();
+    }
```
vigneshsankariyer1234567890, line 123: From this, we only set up a session if `cacheRepoFromUrlService.repoLocation != null`

But setupSession does not use `cacheRepoFromUrlService`, and it will never set up the session if we are using the url encoded repo.

vigneshsankariyer1234567890 (changes requested):
Overall, nice PR and changes look good to go! Have left a couple comments for review, and once those are done, LGTM 🚀

src/app/core/guards/parse-url-params.guard.ts
LJXSean: Hi, i agree with this being an unusual application of angular route guards, but I was not sure of how else to do it.
My original idea was to leave the /issuesViewer path unprotected and inside the ngOnInit() function of IssuesViewerComponent, i would cache the :org/:repo and redirect to login if needed.

But the above approach would require removing authGuard, which didnt seem ideal.

Would you happen to have better alternative approaches?

src/app/shared/layout/header.component.ts
LJXSean: I think gycgabriel mentioned above that this is an existing issue from CATcher. 
When we access pages by manually typing in a new url, it refreshes the whole app. Looking into this, I think we could save the github access token to sessionStorage to prevent re-authentication, but I'm not sure of how secure/safe this would be, considering the scope set for the accessToken

src/app/core/guards/parse-url-params.guard.ts
vigneshsankariyer1234567890: My bad, I was missed a word here (I was supposed to type although definitely *not* wrong). This is completely fine! Sorry about that!

vigneshsankariyer1234567890 (approved):
LGTM, everything looks good to go! 🚀

LJXSean:
Hi @vigneshsankariyer1234567890, if the changes are alright could you help me to merge this PR? 
Thank you!

Issue #229, opened 2024-01-22
App filters do not work with some label names


nknguyenhc:
**Describe the bug**
Specifically, if a label has 2 or more period signs `.`, the filter registers the wrong label name.

**To Reproduce**

1. In a repo of your choice, create a tag with 2 or more period signs `.`, for e.g., `test.type.name`.
2. Attach an issue to the tag.
3. Watch the repo with WATcher.
4. The tags listed in the filter includes `test.type` instead of `test.type.name`.

Alternatively, you can watch my repo at [nknguyenhc/catcher-bugreporting](https://github.com/nknguyenhc/catcher-bugreporting).

**Expected behavior**
The tags listed in the filter should include `test.type.name`.

**Screenshots**

![image](https://github.com/CATcher-org/WATcher/assets/87511888/5fcd31b3-116b-4996-9a7d-49ef0b6e9c97)

**Desktop (please complete the following information):**

- OS: Windows 10
- Browser: chrome
- Version: v1.0.0

**Additional context**
This might be because of how the constructor of a `Label` works. In `label.model.ts`:

```ts
constructor(label: { name: string; color: string; definition?: string }) {
  const containsDotRegex = /\.\b/g; // contains dot in middle of name
  [this.category, this.name] = containsDotRegex.test(label.name) ? label.name.split('.') : [undefined, label.name];
  this.formattedName = this.category === undefined || this.category === '' ? this.name : this.category.concat('.', this.name);
  this.color = label.color;
  this.definition = label.definition;
}
```

The input `label.name` is first split by periods `.`, and then the `formattedName` is only reconstructed from the first two parts of the name.

PR #230, opened 2024-01-22
Fix label filter not working

### LINKED ISSUE #227: Error toast shown on selecting p.Low or priority.Low label on WATcher repository
**Describe the bug**
After selecting p.Low or priority.Low label on WATcher repository, shows this error toast. 

image

nknguyenhc:
### Summary:

Fixes #227 

#### Type of change:

- 🐛 Bug Fix

### Changes Made:

Change the value on the model of `label-filter-bar` component from `label.name` to `label.formattedName`.
The value is used to update the filters on issues and PRs.

The error toast shows because on one click, two filters are applied. I believe this is unintentional.
By changing the value to `label.formattedName`, only one filter is applied (assuming that each label formatted name is unique).

The filter also does not work when the label has 2 parts separated by a period `.`.

Note that the filter still does not work with labels with 2 or more periods `.`. This is because of how `Label`s are constructed. Please see #229 .

### Screenshots:

image

### Proposed Commit Message:

```
Fix label filter not working

Filters takes the value of `label.formattedName` instead of `label.name`.
```

<details><summary>
<h3>Checklist:</h3>
</summary>

</details>

### DIFF
--- a/src/app/shared/filter-bar/label-filter-bar/label-filter-bar.component.html
+++ b/src/app/shared/filter-bar/label-filter-bar/label-filter-bar.component.html
@@ -23,7 +23,7 @@
           <mat-list-option
             #option
             *ngFor="let label of this.labels$ | async"
-            [value]="label.name"
+            [value]="label.formattedName"
             [selected]="selectedLabelNames.includes(label.name)"
             class="list-option"
             [class.hidden]="filter(input.value, label.name)"
--- a/src/app/shared/filter-bar/label-filter-bar/label-filter-bar.component.html
+++ b/src/app/shared/filter-bar/label-filter-bar/label-filter-bar.component.html
@@ -24,20 +24,28 @@
             #option
             *ngFor="let label of this.labels$ | async"
             [value]="label.formattedName"
-            [selected]="selectedLabelNames.includes(label.name)"
+            [selected]="selectedLabelNames.includes(label.formattedName)"
             class="list-option"
             [class.hidden]="filter(input.value, label.name)"
           >
             <div class="flexbox-container">
-              <button mat-icon-button *ngIf="!hiddenLabelNames.has(label.name)" (click)="hide(label.name); $event.stopPropagation()">
+              <button
+                mat-icon-button
+                *ngIf="!hiddenLabelNames.has(label.formattedName)"
+                (click)="hide(label.formattedName); $event.stopPropagation()"
+              >
                 <mat-icon>visibility</mat-icon>
               </button>
-              <button mat-icon-button *ngIf="hiddenLabelNames.has(label.name)" (click)="show(label.name); $event.stopPropagation()">
+              <button
+                mat-icon-button
+                *ngIf="hiddenLabelNames.has(label.formattedName)"
+                (click)="show(label.formattedName); $event.stopPropagation()"
+              >
                 <mat-icon>visibility_off</mat-icon>
               </button>
               <mat-chip
                 [ngStyle]="labelService.setLabelStyle(label.color)"
-                [disabled]="hiddenLabelNames.has(label.name)"
+                [disabled]="hiddenLabelNames.has(label.formattedName)"
                 (click)="simulateClick(option); $event.stopPropagation()"
               >
                 {{ label.formattedName }}


src/app/shared/filter-bar/label-filter-bar/label-filter-bar.component.html
```diff
@@ -23,7 +23,7 @@
           <mat-list-option
             #option
             *ngFor="let label of this.labels$ | async"
-            [value]="label.name"
+            [value]="label.formattedName"
             [selected]="selectedLabelNames.includes(label.name)"
```
gycgabriel, line 27: What about the other uses of label.name instead of formattedName, will there be any edge cases from changing only the value here

nknguyenhc: This `value` being changed to `label.formattedName` only affects the values in `LabelFilterBarComponent::selectedLabelNames`. It is only used internally, to update the value of the behaviour subject field `LabelFilterBarComponent::selectedLabels`. In `LabelFilterBarComponent::simulateClick` and `LabelFilterBar::updateSelection` methods:

```ts
this.selectedLabels.next(this.selectedLabelNames);
```

`LabelFilterBarComponent::updateSelection` value is an `@Input` field. When looking for places that this component is used, it is only used in `FilterBarComponent`, in `filter-bar.component.html`:

```html
<mat-grid-tile class="label-filter-grid-tile" colspan="1">
  <app-label-filter-bar [selectedLabels]="this.labelFilter$" [hiddenLabels]="this.hiddenLabels$"></app-label-filter-bar>
</mat-grid-tile>
```

Hence any changes to `LabelFilterBarComponent::selectedLabels` will only cause a change to `FilterBarComponent::labelFilter$`. Looking at the places that this field is used, it is only used in `FilterBarComponent::ngAfterViewInit`, which subscribes to changes to this field:

```ts
    this.labelFilterSubscription = this.labelFilter$.subscribe((labels) => {
      this.dropdownFilter.labels = labels;
      this.applyDropdownFilter();
    });
```

On every change, the `FilterBarComponent::dropdownFilter.labels` is changed, and then `applyDropdownFilter` is called, where the filter in each of the `views$` are updated to `FilterBarComponent::dropdownFilter`. Here, the `views$` are the child components that are filterable. It is an `@Input` field. Looking at places where `FilterBarComponent` is used, only in `issue-viewer.component.html`:

```html
<app-filter-bar [views$]="views" #filterbar></app-filter-bar>
```

The value of `FilterBarComponent::views$` takes the value of `IssueViewer::views`. `IssueViewer::views` is a field that follows the value of `IssueViewer::cardViews`:

```ts
  ngAfterViewInit(): void {
    this.viewChange = this.cardViews.changes.subscribe((x) => this.views.next(x));
  }
```

Which in turn is a `@ViewChildren` component that is of type `QueryList<CardViewComponent>`! This means that any change to `FilterBarComponent::dropdownFilter` will cause a change to the filters of all `CardViewComponent`s within `IssueViewer`.

TLDR, in short, the change of `value` from `label.name` to `label.formattedName` only means that all `CardViewComponent`s filter by `label.formattedName` instead of `label.name`. Note that `CardViewComponent` has selector `app-view-card`, so to double check where this change will affect, paste the following code to the browser's console when opening the app:

```js
document.querySelectorAll('app-card-view')
```

We see that the `CardViewComponent`s are only the columns of the contributors of the current repo. This is the only component that takes the intended effect of the change.

I realised that the `LabelFilterBarComponent::hiddenLabelNames` needs a similar change. Similar to what I have explained above, the `CardViewComponent`s are the only component that takes the intended effect of the change. However, looks like it only hide the labels. I'm not sure if this is the intended behaviour of the hiding feature, if it not, there needs to be a change to how the `CardViewComponent`s make use of the `filter.hiddenLabels`.

image

After hiding the label `difficulty.Easy`:

image

gycgabriel: > However, looks like it only hide the labels. I'm not sure if this is the intended behaviour of the hiding feature, if it not, there needs to be a change to how the CardViewComponents make use of the filter.hiddenLabels.

To confirm, what did you think the hiding feature should do? We might want to reconsider how we present the features or change the feature if it doesn't make intuitive sense to the user

gycgabriel:
I'm not sure if WATcher even needs to separate the label names anymore, since unlike CATcher we don't have fixed labels and categories for bug reporting phases etc. Can create another issue to refactor Labels, unless I missed out a use case for WATcher.

Thanks for your detailed explanation

gycgabriel (approved):
LGTM

src/app/shared/filter-bar/label-filter-bar/label-filter-bar.component.html
nknguyenhc: I think it should hide the issues & PRs that have the tag, instead of simply hide the tag from the page. I have posted a new issue on this at #240.

Issue #232, opened 2024-01-26
Move Activity Dashboard from prototype to release


gycgabriel:
Currently, WATcher has an Activity Dashboard, but it is hidden as it is not yet ready.

![image](https://github.com/CATcher-org/WATcher/assets/77230723/cb17734d-4d2d-4179-b6ec-f00341566d9d)

Let's bring this feature to WATcher. 

- Improve design of Activity Dashboard to match Issue Viewer
- Hide ID column of events
- Improve Date Selector

damithc:
Let's keep improving the feature, but also keep it hidden until we are sure it is ready.

gycgabriel:
@MadLamprey 
> 1. How is it intended to be accessed - as a button on the Issues Dashboard or some other way?
There was a drop down on the top left for changing from issue to activity, similar to how we change phases in CATcher when we have multiple phases available in development mode. 

> 2. As for the points mentioned in the issue, what has to be improved in the Date selector?

Ircc, current date selector uses two individual date pickers to filter, there is a date picker for a range, but this needs investigation if it is viable to be replaced with that.

PR #233, opened 2024-01-30
Improve activity dashboard design

NereusWB922:
### Summary:

#232 

#### Type of change:

- ✨ Enhancement

### Changes Made:

- Hide the id column of activity dashboard
- Add actor info (avatar and name) to event table
- Make event table height fix and scrollable
- Changed event details display (expanded in table -> dialog)

## Current view

image

## Scrollable

scrollable

## Event details display with dialog

event-details-display

### Proposed Commit Message:

```
Improve activity dashboard design

The ID column in the event table is not useful for the user.

The design of the dashboard is too simple. The event table's height is
not fixed, and the expanded event details might cause the table to be
too long, which is very inconvenient.

To address this, let's hide the ID column, add some information about
the actor, fix the event table height and make it scrollable, and
implement a dialog to display event details.

```

### DIFF
--- a/src/app/activity-dashboard/activity-dashboard.component.css
+++ b/src/app/activity-dashboard/activity-dashboard.component.css
@@ -1,0 +1,4 @@
+.event-tables-wrapper {
+  display: flex;
+  overflow-x: auto;
+}
--- a/src/app/activity-dashboard/activity-dashboard.component.html
+++ b/src/app/activity-dashboard/activity-dashboard.component.html
@@ -28,7 +28,7 @@
   </mat-grid-list>
 </div>
 
-<div style="display: flex">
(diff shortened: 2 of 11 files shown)


damithc:
@NereusWB922 Noticed in passing that the proposed commit message body is not wrapped to the appropriate width.

NereusWB922:
> @NereusWB922 Noticed in passing that the proposed commit message body is not wrapped to the appropriate width.

Updated the commit message. Thank you.

gycgabriel (approved):
LGTM 

I understand this is the starting stages of this dashboard so it's alright, but let's strive to keep future PR sizes smaller
(~100 LoC)

Issue #214, opened 2023-10-09
Ask for public repo access only


damithc:
I don't think most users will be comfortable in granting WATCher wholesale access to their private repos. So, I don't feel we can release this to students' use in this current state.

Suggestion: we should revert back to asking for public repo access only, or at least give users the option to allow or disallow private repo access.

PR #255, opened 2024-02-14
Add shareable repo-specific URL

### LINKED ISSUE #251: Add shareable repo-specific URL
Currently, whenever the user wants to view a repo, user has to login, and then indicate the repo. There is no repo-specific URL that user can share with others. Furthermore, after authentication, the app does not redirect to the pathname initially indicated (if any)

There are 2 parts to this issue:

1. Add repo-specific URL.

This means that a URL in our repo should indicate which repo the intended landing page will be viewing. For example, `https://catcher-org.github.io/WATcher/issueViewer?repo=CATcher-org/WATcher` indicates that the intended landing page will be viewing this repo.

2. Redirect to the intended-landing page.

This means that if the user put `https://catcher-org.github.io/WATcher/issueViewer?repo=CATcher-org/WATcher` to the URL bar in the browser, after authentication, our app should redirect to view the exact same page (i.e. to view this repo), instead of asking from user for which repo to view.

As described below.

**Additional context**
This is to improve user experience, especially first-time user, as discussed in #236. We should try to simplify the process of logging in and viewing a repo for users. Bypassing logging in was considered, however, due to Github's limitation of 60 requests/hr for REST API endpoints and limitation of GraphQL API endpoints to only authenticated user, it was not very feasible.

nknguyenhc:
### Summary:

Fixes #251 

#### Type of change:

- ✨ New Feature/ Enhancement

### Changes Made:

* For every change in repo, the URL query param changes. For example, if you are switching from another repo to this repo, the URL changes to `/issuesViewer?repo=CATcher-org%2FWATcher`.
* When user keys a URL with pathname (i.e. needs to login before proceeding), the pathname + query params are stored in session storage.
* In authentication page, if there is a URL to be redirected to (in session storage), it automates the login process.
* Appropriate error handling is added so that wrong URL input are handled properly.

I did try to limit the changes in this PR to as small as possible, but I could not find a good way to split this task to smaller task, such that smaller tasks can be independently tested.

Some of the sample URLs I have tried:

* `/issuesViewer`: invalid URL, because now issue viewer URL must have query param
* `/issuesViewer?repo=CATcher-org%2FWATcherrr`: invalid URL, because repo does not exist
* `/issuesViewer?repo=CATcher-org%2FWATcher%2Ftest`: invalid URL, because repo name is of wrong format
* `/issuesViewer?repo=CATcher-org%2FWATcher`: valid URL
* Invalid URL, followed by pathname `/`: login should proceed as per normal
* Valid URL, followed by pathname `/`: login should proceed as per normal

### Future enhancements:

Here are some possible future enhancements of this feature:

* Store filters, including labels, milestones, etc, in query params to allow filters to be shareable through links
* Redirect to activity dashboard if URL is indicated as such, when the activity dashboard is available

### Screenshots:

NIL

### Proposed Commit Message:

```
Add shareable repo-specific URL

Previously, URLs are fixed upon repo change, hence URLs are not shareable.

Let's make the current repo visible in the URL, so that it shareable.

Furthermore, login is automated when there is such next route to be redirected to.
```

<details><summary>
<h3>Checklist:</h3>
</summary>

</details>


luminousleek:
Thanks for this PR! A quick test does show that the username doesn't display in the top right corner after logging in though, see the screenshot below.

Screenshot 2024-02-17 at 9 01 14 PM

src/app/core/services/auth.service.ts
```diff
@@ -33,6 +33,9 @@ export enum AuthState {
  * updating the application state with regards to authentication.
  */
 export class AuthService {
+  private static readonly DEFAULT_HAS_PRIVATE_PERMISSION = true;
```
luminousleek, line 36: Not sure if this should be the default. See #214.

nknguyenhc: I have changed this to `false`. This means that auto-redirect does not work if the user paste the link to view a private repo. There needs extra work to handle viewing private repos, I will create a separate issue.

nknguyenhc:
I have fixed the bug! Let me know if there are any other issues.

src/app/core/services/auth.service.ts
```diff
@@ -33,6 +33,9 @@ export enum AuthState {
  * updating the application state with regards to authentication.
  */
 export class AuthService {
+  private static readonly DEFAULT_HAS_PRIVATE_PERMISSION = false;
```
luminousleek, line 36: Small nit - might want to change the constant name here

luminousleek (changes requested):
Looks almost good to merge! Just a really small nit to fix.

src/app/core/services/auth.service.ts
nknguyenhc: I changed the constant name to `DEFAULT_HAS_PERMISSION_TO_PRIVATE_REPOS`

luminousleek: hmm what i meant is that the constant name sort of implies that the constant value is `true`, but it's actually `false`. maybe something like `DEFAULT_DOES_NOT_HAVE_PRIVATE_REPO_PERMISSION` might work

nknguyenhc: To avoid the name being interpreted as a boolean question, I have renamed it to `NO_PERMISSION_TO_PRIVATE_REPOS`

Arif-Khalid: In my opinion I get where @luminousleek is coming from. Default should indicate its default state in the name following the word DEFAULT.
However, this is confusing since developers might mistake it as representing a boolean variable which means "No private repo access", when really it is used as "has private repo access".
I think a better name would be just HAS_PERMISSION_TO_PRIVATE_REPOS. why the DEFAULT at all when it is implied already by static readonly.

Arif-Khalid:
Tested all functionality and working as intended. Aside from naming nitpick by @luminousleek, LGTM!

src/app/core/services/auth.service.ts
nknguyenhc: Without the word `DEFAULT`, it might be mistaken that the auth service will never have access to private repos, while in fact, this is only the default settings when the app skips manual login. The app would still have access to private repos if the user chooses to (with manual login).

Phrasing the name as a boolean question (i.e. `HAS_PERMISSION_TO_PRIVATE_REPOS`) is more applicable when it is a boolean variable that can take up both values (i.e. `true` and `false`) in different scenarios, but in this case, it can only be `false`. Some might infer the variable name as the auth service always have access to private repos.

I would still be more inclined to use `NO_PERMISSION_TO_PRIVATE_REPOS`, because it really means "no private repo access", and is meant to be used only when the app has no private repo access. Unless there is a good way to incorporate the word `DEFAULT` into the constant name when phrasing it as a boolean question.

gycgabriel: Late to the discussion, I would prefer `DEFAULT_NO_PERMISSION_TO_PRIVATE_REPOS` or similar. Default keyword is important to show that it is by default

gycgabriel (commented):
Late to the discussion, I would prefer DEFAULT_NO_PERMISSION_TO_PRIVATE_REPOS or similar. Default keyword is important to show that it is by default

src/app/core/services/auth.service.ts
nknguyenhc: Changed to `DEFAULT_NO_PERMISSION_TO_PRIVATE_REPOS`!

luminousleek (approved):
LGTM, this will be a useful feature! Thanks for your hard work.

Issue #85, opened 2023-02-19
Remove outdated tests


cheehongw:
There are many failing unit tests in WATcher. These tests are inherited from CATcher and require updating.

Reasons for failure fall under these 2 categories:
1. unit under test has been removed when changing from WATcher to CATcher
2. unit under test has undergone changes since CATcher

It should be relatively safe to fix category 1 since we are removing tests for files that don't exist anymore.

Fixes for category 2 should be done carefully, and should try to preserve tests cases as much as possible

<details>
<summary>
<h3>To refactor test:</h3>
</summary>

- [x] Login Component
- [x] Session Model
- [x] Issue Paginator
- [x] Issue Sorter
- [x] Github Label constants
- [x] Search Filter
- [x] Issue Model
- [x] Label Service
- [x] Label Filter Bar Component
- [x] Phase Service
- [x] User Service

</details>

<details>
<summary>
<h3>To remove test:</h3>
</summary>

- [x] markdown.css (in test env stylesheet)
- [x] Conflict Model
- [x] Permissions service
- [x] Unused constants

</details>

chunweii:
Issue is related to #72

chunweii:
@Eclipse-Dominator Are you working on this issue? If not you can unassign yourself and let others work on this

NereusWB922:
Hi, since the outdated tests have been refactored and the pre-push hook for npm run test is also enabled, I think we can close this issue now.

PR #257, opened 2024-02-19
Refactor test cases for Issue Model

NereusWB922:
### Summary:

Fixes part of #85 

#### Type of change:

- 🧪 Tests Update

### Changes Made:

- Refactor `GithubIssue` constants to match current model
- Refactor test cases for issue model

### Proposed Commit Message:

```
Refactor test cases for issue model

The GithubIssue constants and the test cases
for the issue model are outdated. 

Let's refactor the GithubIssue constants and the
test cases for the issue model.
```

### DIFF
--- a/tests/constants/githubissue.constants.ts
+++ b/tests/constants/githubissue.constants.ts
@@ -30,22 +30,19 @@ const randomISODate: (startDate?: Date, endDate?: Date) => string = (
 };
 
 const USER_ANUBHAV_DETAILS = {
-  login: USER_ANUBHAV.loginId,
-  avatar_url: 'https://avatars1.githubusercontent.com/u/35621759?v=4',
-  url: 'https://api.github.com/users/anubh-v'
+  login: USER_ANUBHAV.loginId
 };
 
-const USER_ANUBHAV_ASSIGNEE_DETAILS = {
-  login: USER_ANUBHAV.loginId,
-  id: 35621759,
-  url: 'https://api.github.com/users/anubh-v'
+const MILESTONE_ONE = {
+  number: '1',
+  title: 'Milestone 1',
+  state: 'Open'
 };
 
 export const ISSUE_WITH_EMPTY_DESCRIPTION = new GithubIssue({
   id: '574085971',
   number: 92,
   assignees: [],
-  comments: [],
   body: '',
   created_at: '2020-03-02T16:19:02Z',
   labels: [GITHUB_LABEL_TEAM_LABEL, GITHUB_LABEL_TUTORIAL_LABEL, GITHUB_LABEL_FUNCTIONALITY_BUG, GITHUB_LABEL_MEDIUM_SEVERITY],
@@ -53,15 +50,19 @@ export const ISSUE_WITH_EMPTY_DESCRIPTION = new GithubIssue({
   stateReason: IssueStateReason.Reopened,
   title: 'App starts to lag when given large amount of input',
   updated_at: '2020-03-13T13:37:32Z',
+  closed_at: '',
   url: 'https://api.github.com/repos/CATcher-org/pe-results/issues/92',
-  user: USER_ANUBHAV_DETAILS
+  user: USER_ANUBHAV_DETAILS,
+  milestone: MILESTONE_ONE,
+  comments: [],
+  issueOrPr: 'Issue',
+  isDraft: false
 });
 
 export const ISSUE_UPDATED_EARLIER = new GithubIssue({
   id: '000000001',
   number: 1,
   assignees: [],
-  comments: [],
   body: '',
   created_at: '2020-09-02T16:19:02Z',
   labels: [GITHUB_LABEL_TEAM_LABEL, GITHUB_LABEL_TUTORIAL_LABEL, GITHUB_LABEL_FUNCTIONALITY_BUG, GITHUB_LABEL_MEDIUM_SEVERITY],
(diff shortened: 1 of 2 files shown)


luminousleek:
Currently these tests only test for 4 or so properties of `Issue`? This is probably because the `GithubIssue` constants we're using haven't been adapted for WATcher.

I think we should rewrite the issues in `githubissue.constants.ts` first to match the schema of issues we have in WATcher, and then we refactor these tests.

NereusWB922:
Got it. I will refactor the constants then include check for all properties of issues.

luminousleek (changes requested):
Thanks for this, a minor nit to fix and we should be good!

tests/constants/githubissue.constants.ts
```diff
@@ -29,150 +29,142 @@ const randomISODate: (startDate?: Date, endDate?: Date) => string = (
...
   created_at: '2020-03-02T16:19:02Z',
   labels: [GITHUB_LABEL_TEAM_LABEL, GITHUB_LABEL_TUTORIAL_LABEL, GITHUB_LABEL_FUNCTIONALITY_BUG, GITHUB_LABEL_MEDIUM_SEVERITY],
   state: IssueState.Open,
   stateReason: IssueStateReason.Reopened,
   title: 'App starts to lag when given large amount of input',
   updated_at: '2020-03-13T13:37:32Z',
+  closed_at: '',
   url: 'https://api.github.com/repos/CATcher-org/pe-results/issues/92',
-  user: USER_ANUBHAV_DETAILS
+  user: {
```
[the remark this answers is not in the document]
NereusWB922: That's a good idea. Just pushed the changes. Thank you!

tests/model/issue.model.spec.ts
```diff
@@ -1,143 +1,57 @@
...
-      expect(issue.assignees.length).toBe(1);
+      expect(issue.globalId).toEqual(ISSUE_WITH_EMPTY_DESCRIPTION.id);
+      expect(issue.id).toEqual(ISSUE_WITH_EMPTY_DESCRIPTION.number);
+      expect(issue.created_at).toEqual(moment(ISSUE_WITH_EMPTY_DESCRIPTION.created_at).format('lll'));
+      expect(issue.title).toEqual(ISSUE_WITH_EMPTY_DESCRIPTION.title);
+      expect(issue.description).toEqual('No details provided by bug reporter.');
+      expect(issue.updated_at).toEqual(moment(ISSUE_WITH_EMPTY_DESCRIPTION.updated_at).format('lll'));
+      expect(issue.closed_at).toEqual(moment(ISSUE_WITH_EMPTY_DESCRIPTION.closed_at).format('lll'));
+      expect(issue.milestone).toEqual(
+        ISSUE_WITH_EMPTY_DESCRIPTION.milestone ? new Milestone(ISSUE_WITH_EMPTY_DESCRIPTION.milestone) : Milestone.DefaultMilestone
```
luminousleek, line 18: Sorry didn't catch this earlier but this seems to be the implementation of the method? The expected result should be a new `Milestone` right? It doesn't really make sense to put logic in the expected test result.

And then you can create another test spec with a `GithubIssue` without a milestone to check if the milestone is `Milestone.DefaultMilestone`.

tests/model/issue.model.spec.ts
```diff
@@ -1,143 +1,57 @@
...
-      const dummyTeam = null;
-      const issue = Issue.createPhaseTeamResponseIssue(ISSUE_WITH_ASSIGNEES, dummyTeam);
-      expect(issue.title).toBe('Screen freezes');
-      expect(issue.teamResponse).toBe('No details provided by team.');
-      expect(issue.severity).toBe('Medium');
-      expect(issue.assignees).toContain('anubh-v');
-      expect(issue.assignees.length).toBe(1);
+      expect(issue.globalId).toEqual(ISSUE_WITH_EMPTY_DESCRIPTION.id);
+      expect(issue.id).toEqual(ISSUE_WITH_EMPTY_DESCRIPTION.number);
+      expect(issue.created_at).toEqual(moment(ISSUE_WITH_EMPTY_DESCRIPTION.created_at).format('lll'));
```
luminousleek, line 12: This looks like you're testing the implementation of it? The expected result should be a string in the correct format. How the string is generated shouldn't be what you should be testing for.

tests/model/issue.model.spec.ts
```diff
@@ -1,143 +1,57 @@
...
+        ISSUE_WITH_EMPTY_DESCRIPTION.milestone ? new Milestone(ISSUE_WITH_EMPTY_DESCRIPTION.milestone) : Milestone.DefaultMilestone
+      );
+      expect(issue.state).toEqual(ISSUE_WITH_EMPTY_DESCRIPTION.state);
+      expect(issue.stateReason).toEqual(ISSUE_WITH_EMPTY_DESCRIPTION.stateReason);
+      if (ISSUE_WITH_EMPTY_DESCRIPTION.issueOrPr) {
+        expect(issue.issueOrPr).toEqual(ISSUE_WITH_EMPTY_DESCRIPTION.issueOrPr);
+      }
+      expect(issue.author).toEqual(ISSUE_WITH_EMPTY_DESCRIPTION.user.login);
+      expect(issue.isDraft).toEqual(ISSUE_WITH_EMPTY_DESCRIPTION.isDraft);
+      expect(issue.assignees).toEqual(ISSUE_WITH_EMPTY_DESCRIPTION.assignees.map((assignee) => assignee.login));
```
luminousleek, line 27: You should test this using an issue with assignees. Also, your expected value should be an array of the assignees' logins, rather than using a method to generate it. 

Same thing when testing labels in the line below.

tests/model/issue.model.spec.ts
```diff
@@ -1,143 +1,57 @@
...
     });
   });
-  describe('.updateDescription(description) and .updateTeamResponse(teamResponse)', () => {
+  describe('.updateDescription(description)', () => {
     it('correctly clean strings obtained from users', () => {
       const noDetailsFromBugReporter = 'No details provided by bug reporter.';
-      const noDetailsFromTeam = 'No details provided by team.';
       expect(Issue.updateDescription('')).toBe(noDetailsFromBugReporter);
-      expect(Issue.updateTeamResponse('')).toBe(noDetailsFromTeam);
-      expect(Issue.updateDescription(null)).toBe(noDetailsFromBugReporter);
```
luminousleek, line 36: Not sure why this line is deleted?

luminousleek (changes requested):
Sorry after reviewing another of your PRs I went to take a closer look at this one and it seems that you're testing for implementation rather than the expected behaviour (even if the implementation gives the expected behaviour). Generally the expected values shouldn't have any logic or methods in them. 

Hope this makes sense, if it doesn't you can PM me.

NereusWB922:
> Sorry after reviewing another of your PRs I went to take a closer look at this one and it seems that you're testing for implementation rather than the expected behaviour (even if the implementation gives the expected behaviour). Generally the expected values shouldn't have any logic or methods in them.
> 
> Hope this makes sense, if it doesn't you can PM me.

Thank you so much for the clarification! Will do the changes soon for both prs.

luminousleek (approved):
LGTM, and hope you have a better understanding of how to write tests!

Issue #249, opened 2024-02-08
Refactor filters into its own service


Arif-Khalid:
**Is your feature request related to a problem? Please describe.**
It is frustrating that the filters are declared in multiple components and observables are prop drilled into other components in order to make things work. As a developer working on the filters, it is very difficult to follow.
For example, labelFilter BehaviourSubject is delared in filter-bar, and prop drilled into label-filter-bar.
label-filter-bar is then the component that initializes the labels, calling .next on the behaviour subject from filter-bar.

Additionally, certain filters are referenced objects such as labels being an array and hidden labels being a set. We also use a DEFAULT_DROPDOWN_FILTER variable as the base for when filters are reset. 
Because of these referenced filters, an easy mistake for developers would be these objects' mutation by components, resulting in a permanent mutation of the base DEFAULT_DROPDOWN_FILTER.

This follows from the solution for #205, intended to be worked on after that PR, #226, is merged in.

**Describe the solution you'd like**
A new service filters.service.ts that contains the current repo filters stored as observables and subscribed to the relevant github services such as labelService subscription.
Filters are encapsulated within this component.
Components use the observable's .next and .subscribe functions to access and modify filters

**Describe alternatives you've considered**
We can leave the filters as is and risk errors down the line due to the issues mentioned.
We can use a context provider to lift the state out instead of using a filters service.

* [x] Refactor selected label filter
* [x] Refactor hidden label filter
* [x] Refactor type filter
* [x] Refactor status filter
* [x] Refactor sort filter
* [x] Refactor milestone filter
* [x] Refactor title filter

Arif-Khalid:
@CATcher-org/2223s2 @CATcher-org/2324s2

nknguyenhc:
Agree. I think this one should be done to set the service to be used as solution for #205 (currently #226 is attempting to solve).

What do yall think? @gycgabriel @vigneshsankariyer1234567890

vigneshsankariyer1234567890:
I think it makes a lot of sense to have filters as its own service. That way, we have a neat abstraction to make changes to the filters when needed and access the data within the filters themselves

Arif-Khalid:
All filters have been refactored

PR #259, opened 2024-02-19
Refactor certain filters into its own service

Arif-Khalid:
### Summary:

Fixes part of #249 

#### Type of change:

- 🎨 Code Refactoring

### Changes Made:

* Refactored a few filters into their own service
* Does not break any behaviour for other filters
* Other filters will be refactored in another PR to reduce the size of this one
* Refactored following filters
  * Selected labels
  * Hidden labels
  * Status
  * Type

### Proposed Commit Message:

```
Filters are observed, subscribed and emitted in many different 
components.

This makes the code base hard to understand
and adapt for filtering in future components.

Let's refactor the filters into its own centralised service
```

<details><summary>
<h3>Checklist:</h3>
</summary>

- [x] I have created tests for any new code files created in this PR or provided a link to a issue/PR that addresses this.
- [x] I have updated the project's documentation as necessary.

</details>

### DIFF
--- a/src/app/core/services/filters.service.ts
+++ b/src/app/core/services/filters.service.ts
@@ -1,6 +1,6 @@
 import { Injectable } from '@angular/core';
 import { BehaviorSubject } from 'rxjs';
-import { DEFAULT_DROPDOWN_FILTER, DropdownFilter } from 'src/app/shared/issue-tables/dropdownfilter';
+import { DEFAULT_DROPDOWN_FILTER, DropdownFilter } from '../../shared/issue-tables/dropdownfilter';
 
 @Injectable({
   providedIn: 'root'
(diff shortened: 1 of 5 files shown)


NereusWB922 (commented):
LGTM! Tested the status, type, and label filters, and they are working.

NereusWB922 (approved):
LGTM!

src/app/core/services/filters.service.ts
```diff
@@ -0,0 +1,63 @@
...
+ * Filters are subscribed to and emitted from this service
+ */
+export class FiltersService {
+  public dropdownFilter$ = new BehaviorSubject<DropdownFilter>(DEFAULT_DROPDOWN_FILTER);
+
+  clearFilters(): void {
+    this.dropdownFilter$.next(DEFAULT_DROPDOWN_FILTER);
+  }
+
+  updateSelectedLabels(newSelectedLabels: string[]): void {
```
vigneshsankariyer1234567890, line 20: Seems like there's duplication of the logic when updating dropdownFilter itself. For instance, `updateSelectedLabels` is in essence doing the same thing as `updateHiddenLabels` and even the other methods: to update the right property. Instead, maybe we could follow a pattern of composition using the `Partial<T>` API in TypeScript?

This could look something like this:

```ts
updateFilters(newFilters: Partial<DropdownFilter>): void {
  ...this.dropdownFilter$.value,
  ...newFilters,
}
```

Then after we update, we can check our rules for updating the status and type. 

Consumers would then simply pass some filters that they want to change in the components themselves.

src/app/shared/filter-bar/filter-bar.component.html
```diff
@@ -18,7 +18,7 @@
       </mat-form-field>
       <mat-form-field appearance="standard">
         <mat-label>Type</mat-label>
-        <mat-select [(value)]="this.dropdownFilter.type" (selectionChange)="updateStatusPairing(); applyDropdownFilter()">
+        <mat-select [value]="this.dropdownFilter.type" (selectionChange)="this.filtersService.updateType($event.value)">
```
vigneshsankariyer1234567890, line 21: As per my earlier comment, the change here would be something like

```html
<mat-select [value]="this.dropdownFilter.type" (selectionChange)="this.filtersService.updateFilters({type:$event.value})">
```

vigneshsankariyer1234567890 (changes requested):
So far functionality seems ok! Just a small change requested to the Filters Service :)

src/app/core/services/filters.service.ts
Arif-Khalid: Agreed. This is a type I was not aware of, it is perfect in this situation.
I have made the appropriate changes.

gycgabriel (approved):
Looks neat, thanks for the refactor. Let's wait for vignesh approval

vigneshsankariyer1234567890 (approved):
LGTM 🔥 🚀  changes look great with the addition of the pipe!

PR #253, opened 2024-02-10
Remove unused models

Arif-Khalid:
### Summary:

Fixes part of #72 

#### Type of change:

- 🎨 Code Refactoring

### Changes Made:

- Removed comment.model.ts
- Removed data-file.model.ts
- Removed team,model.ts
- Modified user.model.ts to remove team and allocated team properties

### Proposed Commit Message:

```
Remove unused models in WATcher.

There are a few of unused models ported from CATcher.

Unnecessary models contributes to code clutter and hurts understandability
of code base.

Let's remove these unused models from WATcher to increase code cleanliness.

```

<details><summary>
<h3>Checklist:</h3>
</summary>

- [x] I have tested my changes thoroughly.
- [x] I have created tests for any new code files created in this PR or provided a link to a issue/PR that addresses this.
- [x] I have added or modified code comments to improve code readability where necessary.
- [x] I have updated the project's documentation as necessary.

</details>

PR #262, opened 2024-02-22
Remove constants for DataFile and Team model

NereusWB922:
### Summary:

Fixes part of #85 

#### Type of change:

- 🧪 Tests Update

### Changes Made:

- Remove invalid data constants (models are removed in #253)

### Proposed Commit Message:

```
Remove constants for DataFile and Team model

DataFile and Team model are removed.

Let's remove the constants of these models.
```


luminousleek (approved):
LGTM! I think just rename the PR title to specify which models you're removing.

PR #269, opened 2024-03-02
Fix zone testing import error

### LINKED ISSUE #268: Encounter error for npm run test after upgrading to Angular v11
**Describe the bug**

Encounter error shown below for `npm run test` after upgrading to Angular v11

image

NereusWB922:
### Summary:

Fixes #268 

#### Type of change:

- 🐛 Bug Fix

### Changes Made:

- Move import statement of `zone.js/dist/zone-testing` to the top of `test.ts` file
- Disable tslint rule on ordered-imports for `test.ts` file

### Proposed Commit Message:

```
Fix zone-testing import error

Npm run test will break if the import statement 
for zone-testing is not at the top of test.ts file. 

Let's reorder the import statement and disable
tslint rule for test.ts file.
```

### DIFF
--- a/src/test.ts
+++ b/src/test.ts
@@ -1,8 +1,7 @@
 // This file is required by karma.conf.js and loads recursively all the .spec and framework files
-
+import 'zone.js/dist/zone-testing';
 import { getTestBed } from '@angular/core/testing';
 import { BrowserDynamicTestingModule, platformBrowserDynamicTesting } from '@angular/platform-browser-dynamic/testing';
-import 'zone.js/dist/zone-testing';
 
 declare const require: any;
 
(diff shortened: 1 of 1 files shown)


src/test.ts
```diff
@@ -1,8 +1,8 @@
 // This file is required by karma.conf.js and loads recursively all the .spec and framework files
-
+// tslint:disable:ordered-imports
```
Arif-Khalid, line 2: Developers might find the reason of disabling imports not apparent.
Could you add a comment about the reason this is necessary?

Arif-Khalid (changes requested):
A small and efficient way to disable a particular rule for a particular file. 
Just one small change to documentation.

src/test.ts
NereusWB922: Thanks for the suggestion! Just committed the new changes.

cheehongw (approved):
LGTM

Arif-Khalid (approved):
LGTM!

PR #271, opened 2024-03-02
Refactor test cases for user service

NereusWB922:
### Summary:

Fixes part of #85 

#### Type of change:

- 🧪 Tests Update

### Changes Made:

- Refactor test cases for user service

### Proposed Commit Message:

```
Refactor test cases for user service

The test cases for the user service are outdated. 

Let's refactor test cases for user service.
```

### DIFF
--- a/tests/services/user.service.spec.ts
+++ b/tests/services/user.service.spec.ts
@@ -1,62 +1,35 @@
-import { of } from 'rxjs';
-import { User } from '../../src/app/core/models/user.model';
+import { GithubService } from '../../src/app/core/services/github.service';
+import { LoggingService } from '../../src/app/core/services/logging.service';
 import { UserService } from '../../src/app/core/services/user.service';
-import { jsonData, USER_JUNWEI, USER_Q, USER_SHUMING, USER_WITH_TWO_ROLES } from '../constants/data.constants';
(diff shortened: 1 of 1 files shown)


tests/services/user.service.spec.ts
```diff
@@ -1,62 +1,35 @@
...
 
-    it('should authorize User despite loginId being of different casing', () => {
-      const userService = new UserService(null, dataService);
-      userService.createUserModel(USER_JUNWEI.loginId).subscribe((user) => {
-        expect(user).toBeDefined();
-      });
+  describe('reset()', () => {
+    it('should set current user as undefined', async () => {
+      await userService.createUserModel(USER_JUNWEI.loginId).toPromise();
+      expect(userService.currentUser).toEqual(USER_JUNWEI);
```
vigneshsankariyer1234567890, line 27: Resetting a user to undefined feels weird, although as per the specifications given in User Service, aligns with implementation. Might be worth to change this in a separate issue, to something like null. Null has a meaning which aligns much more closely with what we are trying to achieve here.

vigneshsankariyer1234567890 (approved):
LGTM 🚀 Left a comment on the usage of undefined, we can think about it in a separate issue

Issue #87, opened 2023-02-23
Create tests for WATcher


gycgabriel:
Husky pre-push hooks should be reinstated after tests are created.

File diff: https://github.com/CATcher-org/WATcher/pull/1/files#diff-7ae45ad102eab3b6d7e7896acd08c427a9b25b346470d7bc6507b6481575d519

Github actions should run tests. 

Let's aim to achieve CATcher level of stability eventually.

PR #290, opened 2024-03-11
Create tests for ErrorMessage, ErrorHandling and Milestone services

MadLamprey:
### Summary:

Fixes part of #87

#### Type of change:
- 🧪 Tests Update

### Changes Made:

- Added test cases for `error-handling.service.ts`, `error-message.service.ts` and `milestone.service.ts`

### Proposed Commit Message:

```
Create tests for ErrorMessage, ErrorHandling and Milestone services

Currently, there are no tests written for ErrorMessage, ErrorHandling and Milestone services.

Let's create tests for ErrorMessage, ErrorHandling and Milestone services.
```


tests/services/error-message.service.spec.ts
```diff
@@ -0,0 +1,117 @@
...
+    it('should return the correct message', () => {
+      expect(ErrorMessageService.multipleDropdownOptionsErrorMessage()).toBe('Supply either Dropdown option number or text, not both.');
+    });
+  });
+
+  describe('ErrorMessageService: noDropdownOptionsErrorMessage()', () => {
+    it('should return the correct message', () => {
+      expect(ErrorMessageService.noDropdownOptionsErrorMessage()).toBe('No Dropdown identification parameters supplied.');
+    });
+  });
```
NereusWB922, line 116: I'm not sure if there's any benefit to write test cases for the constant provider service. If test cases fail due to future changes in constant values, developers will likely just edit the test cases, resulting in duplicated changes. Also, I think that integration tests for services using these constants should suffice to catch any errors in those constant values.
@CATcher-org/2324s2  what do yall think?

luminousleek (changes requested):
Thanks for this PR! Some small changes needed, and as per Nereus's comment I don't think we need tests for `ErrorMessageService`. 

Also, please split this PR into one PR for each service you are writing tests for.

tests/services/milestone.service.spec.ts
```diff
@@ -0,0 +1,52 @@
...
+    it('should fetch all milestones', async () => {
+      const mockMilestones = [{ title: 'Milestone 1' }, { title: 'Milestone 2' }];
+      githubServiceSpy.fetchAllMilestones.and.returnValue(of(mockMilestones));
+      milestoneService.fetchMilestones().subscribe((response) => {
+        expect(githubServiceSpy.fetchAllMilestones).toHaveBeenCalled();
+        expect(response).toEqual(mockMilestones);
+        expect(milestoneService.milestones.length).toBe(3);
+        expect(milestoneService.milestones[0].title).toBe('Milestone 1');
+        expect(milestoneService.hasNoMilestones).toBeFalse();
+      });
```
NereusWB922, line 25: Consider using Jasmine's `done()` callback at the end of the assertions. This is to ensure that the test doesn't complete before the value is emitted, which would lead to the assertions not being executed by Jasmine.

The same issue applies to the test case below.

NereusWB922 (changes requested):
Great work on adding new test cases! Just a minor suggestion to ensure proper handling of asynchronous behavior in the test.

tests/services/milestone.service.spec.ts
MadLamprey: I looked into this, and received an Error Log.

Screenshot 2024-03-15 at 2 11 33 PM

NereusWB922: If I am not wrong, async/await cannot be used with done. You try remove the async keyword.

MadLamprey:
PR has been separated into #303 and #304.

PR #314, opened 2024-03-24
Add filters to url

### LINKED ISSUE #277: Add current filters to URL
As a follow up from #251, we should include filters into URL, and allow the filters to be shareable through URL as well.

Add a query param to indicate filters for milestones, sort ID, etc, and apply the corresponding param upon landing onto the same URL, similar to #255 .

NIL

**Additional context**
As a follow up of #251.

Arif-Khalid:
### Summary:

Fixes #277 

#### Type of change:

- ✨ New Feature/ Enhancement

### Changes Made:

- Add current filters to URL
- Filters are set from URL in ``IssuesViewer::ngOnInit``
- Filters are pushed to url (without routing) in filter service whenever they are updated
- Preset views are also stored and retrieved from URL

### Proposed Commit Message:

```
Filters cannot be shared to among users.

Users might want to share their current view
to others, consisting of their current filters.

Let's pull and store filters in the URL to allow
sharing of filter combinations.
```

<details><summary>
<h3>Checklist:</h3>
</summary>

</details>


src/app/core/services/label.service.ts
```diff
@@ -28,7 +28,7 @@ export class LabelService {
   simpleLabels: SimpleLabel[];
 
   private labelsPollSubscription: Subscription;
-  private labelsSubject = new BehaviorSubject<SimpleLabel[]>([]);
+  private labelsSubject = new Subject<SimpleLabel[]>();
```
Arif-Khalid, line 31: The reason for changing ``labelsSubject'' from type ''BehaviourSubject'' to ''Subject'' is because ''BehaviourSubject'' emits an initial empty array as the all labels set. 
As a result, when opening a repo through a url with some specified labels in the url, those labels are sanitized immediately because of the initial emission of the empty all labels.
The change to ''Subject'' removes this initial empty labels emission, solving the issue.

nknguyenhc (commented):
Please update your branch. We are no longer using `_validateFilters` in filters service.

Some preliminary comments on the filter:
* It does not work when I paste a URL containing the filters with milestones into the browser. It defaults back to selecting all milestones.
* Filters with one-value can be made more readable by removing the quotation marks around. For e.g. filter by type can be encoded as `type=issue` instead of `type="issue"`

Arif-Khalid:
> Please update your branch. We are no longer using `_validateFilters` in filters service.
> 
> Some preliminary comments on the filter:
> 
> * It does not work when I paste a URL containing the filters with milestones into the browser. It defaults back to selecting all milestones.
> * Filters with one-value can be made more readable by removing the quotation marks around. For e.g. filter by type can be encoded as `type=issue` instead of `type="issue"`

This PR was created before milestones were saved hence the behaviour. 
I agree with the encoding behaviour suggested and would also like to use the same method of subscribing to pop state events precedented in #319.
Will update this PR with the appropriate behaviour once changes to FiltersService are finalised in the pending PRs.
Converting to draft for the time being.

src/app/core/services/filters.service.ts
```diff
@@ -61,21 +64,77 @@ export class FiltersService {
...
+    const nextFilter: Filter = this.defaultFilter();
+    const queryParams = this.route.snapshot.queryParamMap;
+    try {
+      const presetView = queryParams.get(FiltersService.PRESET_VIEW_QUERY_PARAM_KEY);
+
+      // Use preset view if set in url
+      if (presetView !== 'custom') {
+        this.updatePresetView(presetView);
+        return;
+      }
```
NereusWB922, line 109: `null` value and invalid `presetView` are not handled. We can simply handle them by calling `updatePresetView` with the default value.

NereusWB922 (changes requested):
Just a minor issue on presetView param handling in initialize method.

src/app/core/services/filters.service.ts
Arif-Khalid: Agreed. I have made the appropriate changes

NereusWB922 (approved):
LGTM

PR #315, opened 2024-03-25
Split 'Without a milestone' option

### LINKED ISSUE #297: Split 'Without a milestone' option
Current:

image

Suggestion: Split option as 
```
...
[ ] PRs without a milestone
[ ] Issues without a milestone
```

Reason: The first category is of high interest (likely these are PRs the team forgot to assign to a milestone) while the second category is of low interest.

MadLamprey:
### Summary:

Fixes #297 

### Changes Made:

- Alter the Milestones dropdown to have `Issue without a milestone` and `PR without a milestone` instead of `Without a milestone` when `type` is selected as `All`
- Github issues without a milestone have `IssueWithoutMilestone` or `PRWithoutMilestone` depending on their type, instead of `DefaultMilestone`

### Screenshots:
Screenshot 2024-03-25 at 1 07 56 PM
Screenshot 2024-03-25 at 1 08 12 PM

### Proposed Commit Message:

```
Split `Without a milestone` option

Currently, there is only a `Without a milestone` option even when the type is selected as `All`.
It will be useful if this can be split into `Issue without a milestone` and `PR without a milestone`.

Let's split the `Without a milestone` option to include the two options.
```


src/app/shared/issue-tables/dropdownfilter.ts
```diff
@@ -1,3 +1,4 @@
+import { milestone } from '@primer/octicons';
```
nknguyenhc, line 1: Unused import

src/app/core/services/milestone.service.ts
```diff
@@ -45,6 +49,9 @@ export class MilestoneService {
 
     // add default milestone for untracked issues/PRs at the end
     milestoneData.push(Milestone.DefaultMilestone);
+    milestoneData.push(Milestone.IssueWithoutMilestone);
+    milestoneData.push(Milestone.PRWithoutMilestone);
+
```
nknguyenhc, line 54: Since we are splitting having no milestones into issues and PRs, should we deprecate `Milestone.DefaultMilestone`?

src/app/core/models/milestone.model.ts
```diff
@@ -3,6 +3,8 @@
  */
 export class Milestone {
   static DefaultMilestone: Milestone = new Milestone({ title: 'Without a milestone', state: null });
+  static IssueWithoutMilestone: Milestone = new Milestone({ title: 'Issue without a milestone', state: null });
+  static PRWithoutMilestone: Milestone = new Milestone({ title: 'PR without a milestone', state: null });
```
nknguyenhc, line 7: Should the title be `Issues` and `PRs`, each with an 's', instead?

nknguyenhc (changes requested):
image

My TP repo has some issues unassigned to any milestone. However, the following steps would otherwise shows that there are no issue unassigned:

1. Make sure that your `Type` filter is `Pull Request`.
2. Select only `Without a milestone`.
3. Select `Type` filter to `Issue`.

image

In my opinion, we should deprecate the `Without a milestone` option totally, so that to eliminate this kind of bug.

src/app/core/services/milestone.service.ts
```diff
@@ -24,7 +25,10 @@ export class MilestoneService {
   public fetchMilestones(): Observable<any> {
     return this.githubService.fetchAllMilestones().pipe(
       map((response) => {
-        this.milestones = this.parseMilestoneData(response);
+        const parsedMilestones = this.parseMilestoneData(response);
+        this.milestones = [...parsedMilestones];
+        this.milestones.splice(-3, 1);
+        this.altMilestones = parsedMilestones.slice(0, -2);
         this.hasNoMilestones = response.length === 0;
```
nknguyenhc, line 32: Not sure why there needs to be this logic. Since we only want the actual milestones and not the default milestones, why not just do not append the two default milestones in `parseMilestoneData` at all? Instead of appending them and then slicing them off here.

src/app/shared/filter-bar/filter-bar.component.html
```diff
@@ -70,7 +70,7 @@
           <mat-select-trigger *ngIf="this.milestoneService.hasNoMilestones">
             <span>No Milestones</span>
           </mat-select-trigger>
-          <mat-option *ngFor="let milestone of this.milestoneService.milestones" [value]="milestone.title">
+          <mat-option *ngFor="let milestone of getDisplayedMilestones()" [value]="milestone.title">
```
nknguyenhc, line 73: Not really sure if creating a method to obtain the available list of milestones is really necessary. I think we can make use of `ngIf` directive to conditionally render the default milestone options? See `filter-bar.component.html`, lines 27-31 to see how the directive is used to conditionally render the status checkboxes.

nknguyenhc: Additionally, this also allows us to differ the text rendered from the value, i.e. the texts rendered should be `Issues without a milestone` and `PRs without a milestone`, the nouns with an `s`.

src/app/shared/issue-tables/dropdownfilter.ts
```diff
@@ -38,6 +38,14 @@ export function applyDropdownFilter(filter: Filter, data: Issue[]): Issue[] {
...
     }
 
+    if (filter.milestones.some((milestone) => issue.milestone.title === 'Issue without a milestone')) {
+      ret = ret && issue.issueOrPr === 'Issue';
+    }
+
+    if (filter.milestones.some((milestone) => issue.milestone.title === 'PR without a milestone')) {
+      ret = ret && issue.issueOrPr === 'PullRequest';
+    }
+
```
nknguyenhc, line 48: These lines are not really necessary. The following lines with filtering using `filter.milestones` is already sufficient to also filter in issues and PRs without milestones.

src/app/core/services/milestone.service.ts
nknguyenhc: Also, not really sure what the `altMilestones` is for, other than this assignment of value, it is not used elsewhere.

nknguyenhc (changes requested):

nknguyenhc (commented):
Please resolve conflict with main branch. The conflict is causing the filters on milestones to not be kept properly when switching repos.

nknguyenhc:
image

nknguyenhc (changes requested):
An overlooked bug from a previous PR,

1. View the repo `AY2324S2-CS2103T-T08-2/tp`
2. Select milestones `v1.1` and `Issues without a milestone`
3. Switch to `AY2324S2-CS2103T-T08-1/tp` with keep filters option

Expected result: filter with milestones `v1.1` and `Issues without a milestone` should be used.

Actual result: filter is default back to all milestones.

Context: in `filters.service.ts`, lines 83-103:

```ts
    if (this.filter$.value.milestones.length === this.previousMilestonesLength) {
      this.updateFilters({ milestones: [...allMilestonesSet] });
      this.previousMilestonesLength = allMilestones.length;
      return;
    }
    const newMilestones: string[] = [];
    for (const milestone of this.filter$.value.milestones) {
      if (allMilestonesSet.has(milestone)) {
        newMilestones.push(milestone);
      }
    }
    // No applicable milestones, reset to all milestones selected
    if (newMilestones.length === 0) {
      newMilestones.push(...allMilestonesSet);
    }
    this.updateFilters({ milestones: newMilestones });
    this.previousMilestonesLength = allMilestones.length;
```

`this.previousMilestonesLength` is assigned to `allMilestones.length`, which excludes the default milestones. Instead, it should be assigned to `allMilestonesSet.size`, which includes the default milestones.

nknguyenhc (approved):
LGTM! There is another assignment of `this.previousMilestoneLength` that I have also updated.

nknguyenhc (approved):

PR #319, opened 2024-03-26
Include groupby params in url

### LINKED ISSUE #317: Encode grouping to url
Is your feature request related to a problem? Please describe.
We should include grouping into URL, so the grouping can be shareable as well.

Describe the solution you'd like
Add a query param to indicate grouping. e.g. `?groupby=milestone`

Describe alternatives you've considered
NIL

NereusWB922:
### Summary:

Fixes #317

#### Type of change:

(Delete where appropriate)

- ✨ New Feature/ Enhancement

### Changes Made:

- Update `setCurrentGroupingType` method: Allow setting the `replaceUrl` field
  - True: Replace the current url state in history
  - False: Push a new url state to history
- Implement initialization with url for `GroupingContextService`
   - Initialize when init `IssueViewer` page
   - Initialize when clicking back button or forward button ("popstate")
- Reset `GroupingContextService` if not keeping filter when switching repo
- Preserve previous url param when changing repo (`ViewService`)
  - Include the field `queryParamHandling: 'merge'`

### Proposed Commit Message:

```
Include groupby parameters in URL

With groupby parameters in the URL, users can easily share 
the current grouping by using the URL.

Let's implement initialization with URL parameters and 
the ability to set groupby parameters.
```

### DIFF
--- a/src/app/core/services/grouping/grouping-context.service.ts
+++ b/src/app/core/services/grouping/grouping-context.service.ts
@@ -47,19 +47,17 @@ export class GroupingContextService {
     const groupByParam = this.route.snapshot.queryParamMap.get(GroupingContextService.GROUP_BY_QUERY_PARAM_KEY);
 
     if (groupByParam && Object.values(GroupBy).includes(groupByParam as GroupBy)) {
-      this.setCurrentGroupingType(groupByParam as GroupBy, true);
+      this.setCurrentGroupingType(groupByParam as GroupBy);
     } else {
-      this.setCurrentGroupingType(DEFAULT_GROUPBY, true);
+      this.setCurrentGroupingType(DEFAULT_GROUPBY);
     }
   }
 
   /**
    * Sets the current grouping type and updates the corresponding query parameter in the URL.
    * @param groupBy The grouping type to set.
-   * @param replaceUrl Determines whether to replace the current URL in the browser's history.
-   *                   If true, it replaces the URL; if false, it adds a new entry to the history.
    */
-  setCurrentGroupingType(groupBy: GroupBy, replaceUrl: boolean): void {
+  setCurrentGroupingType(groupBy: GroupBy): void {
     this.currGroupBy = groupBy;
     this.currGroupBySubject.next(this.currGroupBy);
 
@@ -69,7 +67,7 @@ export class GroupingContextService {
         [GroupingContextService.GROUP_BY_QUERY_PARAM_KEY]: groupBy
       },
       queryParamsHandling: 'merge',
-      replaceUrl: replaceUrl
+      replaceUrl: true
     });
   }
 
@@ -107,6 +105,6 @@ export class GroupingContextService {
    * Resets the current grouping type to the default.
    */
   reset(): void {
-    this.setCurrentGroupingType(DEFAULT_GROUPBY, false);
+    this.setCurrentGroupingType(DEFAULT_GROUPBY);
   }
 }
--- a/src/app/core/services/grouping/grouping-context.service.ts
+++ b/src/app/core/services/grouping/grouping-context.service.ts
@@ -1,4 +1,5 @@
 import { Injectable, Injector } from '@angular/core';
+import { ActivatedRoute, Router } from '@angular/router';
 import { BehaviorSubject, Observable } from 'rxjs';
 import { Group } from '../../models/github/group.interface';
 import { Issue } from '../../models/issue.model';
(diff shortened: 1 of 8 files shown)


src/app/issues-viewer/issues-viewer.component.ts
```diff
@@ -55,10 +62,23 @@ export class IssuesViewerComponent implements OnInit, AfterViewInit, OnDestroy {
...
+      .pipe(filter((event) => event instanceof NavigationEnd || event instanceof NavigationStart))
+      .subscribe((event) => {
+        if (event instanceof NavigationStart && event.navigationTrigger === 'popstate') {
+          this.popStateNavigationId = event.id;
+        }
+
+        if (event instanceof NavigationEnd && event.id === this.popStateNavigationId) {
+          this.groupingContextService.initializeFromUrlParams();
+        }
+      });
```
Arif-Khalid, line 76: Since we already have a ``ViewService::setupFromURL`` function that is only called when a user loads into WATcher via url, why not trigger initialize from URL there?
To keep the url loading functionality in the same place, I think the function should be called there instead of subscribing to the popStateEvent. 
This will also mean we don't arbitrarily load from URL the same groupBy that is already in grouping context whenever we route, instead only loading from URL on initial setup.

src/app/core/services/grouping/grouping-context.service.ts
```diff
@@ -36,12 +38,36 @@ export class GroupingContextService {
...
+    }
+  }
+
+  /**
+   * Sets the current grouping type and updates the corresponding query parameter in the URL.
+   * @param groupBy The grouping type to set.
+   * @param replaceUrl Determines whether to replace the current URL in the browser's history.
+   *                   If true, it replaces the URL; if false, it adds a new entry to the history.
+   */
+  setCurrentGroupingType(groupBy: GroupBy, replaceUrl: boolean): void {
```
Arif-Khalid, line 59: Is there a necessity for specifying replaceURL for group by?
In my opinion, it is unnatural for me to be changing back to prior group by if I press the previous page button on my browser, I would instead expect to go to a previous page.
It is a matter of UX and should be discussed further, if others agree then we should always set replaceURL to true and remove this parameter to prevent polluting the user history with unintuitive routes specifying different group bys.

src/app/core/services/view.service.ts
```diff
@@ -88,7 +88,8 @@ export class ViewService {
     this.router.navigate(['issuesViewer'], {
       queryParams: {
         [ViewService.REPO_QUERY_PARAM_KEY]: repo.toString()
-      }
+      },
+      queryParamsHandling: 'merge'
```
Arif-Khalid, line 92: Excellent spot of a source of potential bugs with the new additions

Arif-Khalid (commented):
Overall an excellent PR which effectively establishes a valued feature.
I would recommend reducing unnecessary code by reusing already established functionality and consider the UX when handling the user browser history, although it is just my opinion.

src/app/issues-viewer/issues-viewer.component.ts
NereusWB922: In my opinion, the ViewService is responsible for handling the current view and current repository. I don't think it would be suitable to set up the grouping service within the ViewService. What if we have a view that doesn't utilize the grouping service?

The reason for subscribing to the popStateEvent is to ensure that when navigating back to the previous issue viewer page, we update our grouping service with the "group by" parameter from the previous page. One scenario is that I set the "group by" value to "Milestone" for the current repository, then I switch to another repository without maintaining the filter. When I click the back button, I should expect to return to the previous repository with the previous "group by" and filters.

Arif-Khalid: I see and I am inclined to agree with this behaviour. In that case an excellent use of the pop state event.

Arif-Khalid (approved):
Current implementation LGTM!

src/app/core/services/grouping/grouping-context.service.ts
NereusWB922: Just fixed it to always set 'replaceUrl' to true.

NereusWB922:
Some clarifications for the new commits:

1. Now call the `reset` method of `GroupingContextService` after navigating to a new repo. This ensures that the `reset` method won't replace the `groupby` params of the previous repo.
2. Fixed some bugs caused by `Subscription` and `Timeout`. In progress Timeout should be cleared and Subscription should be unsubscribed in `onDestroy` to avoid memory leaks and prevent unexpected behavior."

Arif-Khalid (approved):
LGTM!

PR #318, opened 2024-03-25
Add sorting by Status

MadLamprey:
### Summary:

Fixes #301 

### Changes Made:

- Add functionality to sort by Status
- The order that is followed is: Open Pull Request > Open Issue > Merged Pull Request > Closed Issue > Closed Pull Request
- For issues/PRs having the same status, they are sorted alphabetically

### Screenshots:
Screenshot 2024-03-25 at 11 19 08 PM

### Proposed Commit Message:

```
Add sorting by Status

Currently, there is no way to sort by Status.
This can be a useful feature for the users to give a notion of priority.

Let's add functionality to sort by Status.
```

PR #320, opened 2024-03-27
Add preset views

### LINKED ISSUE #300: Support preset views
Currently, we have ways for the user to finetune each aspect  of the issues dashboard view.
image

Suggestion: provide an easy way for users to reach commonly used preset views without having to hand-tune multiple aspect.

For example, a view I think most users will want to see the following (this could even be the default view):

`currently active`: all issues and PRs except closed PRs, assigned to the currently active milestone, including PRs (but not issues) that do not have a milestone, sorted by status (i.e., open issues/PRs on top)

What are other such presets that could be useful?
* `milestone X` all issues and PRs assigned to milestone X, sorted by ID(?)
* `contribution of JohnDoe`, showing what the person contributed to each milestone. But this is entirely different view from the current dashboard layout i.e., one column per each milestone

nknguyenhc:
### Summary:

Fixes #300

#### Type of change:

- ✨ New Feature/ Enhancement

### Changes Made:

- Add deadline field to milestone.
- Add preset views.
- Apply preset views on filters.
- Change header to contain a button to select preset view.

"Currently active" preset view depends on sorting by status, so changes in this PR will fully work when #318 is merged.

Any changes made to type, status and milestones filters will set the preset view value to "custom".

### Screenshots:

image

### Proposed Commit Message:

```
Add preset views

Currently, there are a lot of options of filters,
but there are no suggested useful preset views.

We add preset views for users to choose from,
and via a button on the header.
```

<details><summary>
<h3>Checklist:</h3>
</summary>

</details>

### DIFF
--- a/src/app/core/services/filters.service.ts
+++ b/src/app/core/services/filters.service.ts
@@ -16,17 +16,6 @@ export type Filter = {
   deselectedLabels: Set<string>;
 };
 
-export const DEFAULT_FILTER: Filter = {
-  title: '',
-  status: ['open pullrequest', 'merged pullrequest', 'open issue', 'closed issue'],
-  type: 'all',
-  sort: { active: 'id', direction: 'asc' },
-  labels: [],
-  milestones: [],
-  hiddenLabels: new Set<string>(),
-  deselectedLabels: new Set<string>()
-};
-
 @Injectable({
   providedIn: 'root'
 })
@@ -62,11 +51,12 @@ export class FiltersService {
   };
 
   // List of keys in the new filter change that causes current filter to not qualify to be a preset view.
-  readonly presetChangingKeys = new Set<string>(['status', 'type', 'milestones']);
+  readonly presetChangingKeys = new Set<string>(['status', 'type', 'milestones', 'labels', 'hiddenLabels', 'deselectedLabels']);
 
-  public filter$ = new BehaviorSubject<Filter>(DEFAULT_FILTER);
+  readonly defaultFilter = this.presetViews.currentlyActive;
+  public filter$ = new BehaviorSubject<Filter>(this.defaultFilter());
   // Either 'currentlyActive', 'contributions', or 'custom'.
-  public presetView$ = new BehaviorSubject<String>('custom');
+  public presetView$ = new BehaviorSubject<string>('currentlyActive');
 
   // Helps in determining whether all milestones were selected from previous repo during sanitization of milestones
   private previousMilestonesLength = 0;
@@ -74,8 +64,8 @@ export class FiltersService {
   constructor(private milestoneService: MilestoneService) {}
 
   clearFilters(): void {
-    this.filter$.next(DEFAULT_FILTER);
-    this.presetView$.next('custom');
+    this.filter$.next(this.defaultFilter());
+    this.presetView$.next('currentlyActive');
     this.previousMilestonesLength = 0;
   }
 
@@ -86,6 +76,21 @@ export class FiltersService {
     };
     this.filter$.next(nextDropdownFilter);
     this.updatePresetViewFromFilters(newFilters);
+  }
+
+  /**
+   * Updates the filters without updating the preset view.
+   * This should only be called when there are new labels/milestones.
+   * The preset view will be reapplied.
+   * @param newFilters The filters with new values
+   */
+  private updateFiltersWithoutUpdatingPresetView(newFilters: Partial<Filter>): void {
+    const nextDropdownFilter: Filter = {
+      ...this.filter$.value,
+      ...newFilters
+    };
+    this.filter$.next(nextDropdownFilter);
+    this.filter$.next(this.presetViews[this.presetView$.value]());
   }
 
   private updatePresetViewFromFilters(newFilter: Partial<Filter>): void {
@@ -125,7 +130,11 @@ export class FiltersService {
 
     const newLabels = this.filter$.value.labels.filter((label) => allLabelsSet.has(label));
 
-    this.updateFilters({ labels: newLabels, hiddenLabels: newHiddenLabels, deselectedLabels: newDeselectedLabels });
+    this.updateFiltersWithoutUpdatingPresetView({
+      labels: newLabels,
+      hiddenLabels: newHiddenLabels,
+      deselectedLabels: newDeselectedLabels
+    });
   }
 
   sanitizeMilestones(allMilestones: Milestone[]) {
@@ -133,7 +142,7 @@ export class FiltersService {
 
     // All previous milestones were selected, reset to all new milestones selected
     if (this.filter$.value.milestones.length === this.previousMilestonesLength) {
-      this.updateFilters({ milestones: [...allMilestonesSet] });
+      this.updateFiltersWithoutUpdatingPresetView({ milestones: [...allMilestonesSet] });
       this.previousMilestonesLength = allMilestones.length;
       return;
     }
@@ -150,7 +159,7 @@ export class FiltersService {
       newMilestones.push(...allMilestonesSet);
     }
 
-    this.updateFilters({ milestones: newMilestones });
+    this.updateFiltersWithoutUpdatingPresetView({ milestones: newMilestones });
     this.previousMilestonesLength = allMilestones.length;
   }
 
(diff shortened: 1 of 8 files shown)


NereusWB922:
In current implementation, seems like the preset view only include preset value for filters. What if we need preset view with different grouping in the future? Should we include groupby value in the preset value?

nknguyenhc:
> In current implementation, seems like the preset view only include preset value for filters. What if we need preset view with different grouping in the future? Should we include groupby value in the preset value?

Agree, but given the time constraint until the next release, I would be more inclined to just focus on the filters first, and then let this preset view handle the groupby feature later. Also since logic for groupby is in a separate service, I would create another PR for the additional logic.

src/app/core/services/filters.service.ts
```diff
@@ -34,13 +35,47 @@ export const DEFAULT_FILTER: Filter = {
...
+      status: ['open pullrequest', 'merged pullrequest', 'open issue', 'closed issue'],
+      type: 'all',
+      sort: { active: 'id', direction: 'desc' },
+      labels: [],
+      milestones: this.milestoneService.milestones.map((milestone) => milestone.title),
+      hiddenLabels: new Set<string>(),
+      deselectedLabels: new Set<string>()
+    }),
+    custom: () => this.filter$.value
+  };
```
Arif-Khalid, line 62: Since changing labels, hiddenLabels, deselectedLabels, sort and title do not make the preset view "custom", should we reset them or set them to a fixed value when setting the preset view?
To me, it is unintuitive that since it is "not part of the preset views" since changing these parameters did not affect the status of my preset view being put into custom, choosing a preset view should not reset these parameters.

Arif-Khalid: Additionally, is it intentional to provide currentlyActive preset with an invalid sort id of status?

src/app/core/services/filters.service.ts
```diff
@@ -99,4 +153,18 @@ export class FiltersService {
...
+      return [earliestOpenMilestone];
+    }
+
+    const latestClosedMilestone = this.milestoneService.getLatestClosedMilestone();
+    if (latestClosedMilestone) {
+      return [latestClosedMilestone];
+    }
+
+    return this.milestoneService.milestones;
+  }
```
Arif-Khalid, line 169: Good handling of edge cases

src/app/core/services/milestone.service.ts
```diff
@@ -47,4 +47,42 @@ export class MilestoneService {
...
+        continue;
+      }
+      if (latestClosedMilestone === null) {
+        latestClosedMilestone = milestone;
+      } else if (milestone.deadline > latestClosedMilestone.deadline) {
+        latestClosedMilestone = milestone;
+      }
+    }
+    return latestClosedMilestone;
+  }
```
Arif-Khalid, line 87: Functions are well written, documented and accomplishes its task in a simple way. 
Perfect representation of KISS.

src/app/shared/layout/header.component.html
```diff
@@ -12,10 +12,10 @@
     >WATcher v{{ this.getVersion() }}</a
   >
   <span id="view-descriptor" *ngIf="auth.isAuthenticated()" style="margin-left: 10px">
-    ({{ this.getViewDescription(viewService.currentView) }})
+    ({{ this.presetViews[this.filtersService.presetView$.value] }})
   </span>
 
-  <div *ngIf="auth.isAuthenticated() && this.viewService.sessionData.sessionRepo.length > 1">
+  <!--div *ngIf="auth.isAuthenticated() && this.viewService.sessionData.sessionRepo.length > 1">
```
Arif-Khalid, line 18: Should this redundant code be left in the code base?

Arif-Khalid (commented):
A great quality PR which highlights a few good SWE principles.
Some comments about the behaviour of preset views in relation to changing filters as well as a possible error in the preset view filters.
Overall a good PR, well done!

src/app/shared/layout/header.component.html
nknguyenhc: We are still unsure whether to keep the activity dashboard. Currently, this is the only gateway to the activity dashboard, so I'm afraid it will be more difficult for the next batch of ppl to unhide activity dashboard, if we ever want to develop it.

Arif-Khalid: I see. In that case could you add a comment highlighting its significance for other developers who might remove it on instinct?

src/app/core/services/filters.service.ts
nknguyenhc: I would be more inclined to standardise the preset views. The purpose of "currently active" preset view, for e.g., is to get the distribution of work between people within the current milestone. The user might forget to clear all label filters before setting the preset view to "currently active", and thereby get the wrong impression of the work distribution.

src/app/core/services/filters.service.ts
```diff
@@ -34,13 +35,47 @@ export const DEFAULT_FILTER: Filter = {
...
   // Helps in determining whether all milestones were selected from previous repo during sanitization of milestones
   private previousMilestonesLength = 0;
 
+  constructor(private milestoneService: MilestoneService) {}
+
   clearFilters(): void {
     this.filter$.next(DEFAULT_FILTER);
+    this.presetView$.next('custom');
     this.previousMilestonesLength = 0;
   }
```
NereusWB922, line 80: Similar comment as previous one, should we have "Currently Active" as our default view?

NereusWB922 (changes requested):
Just a small suggestion on the default view.

src/app/core/services/filters.service.ts
Arif-Khalid: I see and that makes sense. Is there a reason the other filters aside from the labels are not "tracked" by the preset views?

nknguyenhc: I would argue that the more important part of the preset view is to get the total count. So for the other filters not being tracked:

* Title: the user might want to search for a particular issue/PR. Technically, the user still stays in the same preset view, as the user is still in the same preset view when the search input is cleared.
* Sort: arguably, sort is an important part of the filters, especially for currently active, but user still can choose their own sorting order, and the total count of each column does not change.
* Labels, selected labels, deselected labels: as explained above, we can look into including these as one of the "tracked" filters.
* Hidden labels: This only affects which label is shown on each PR/issue card and which is not, so total count of each column does not change.

These are just my reasoning, let me know if you think otherwise.

Arif-Khalid: Your reasoning is sound. I am inclined to agree but still think there may be some confusion.
When I was testing it, I found it strange that my preset views were maintained when i did something like search by title yet my title was reset when i chose a preset view.
Perhaps a new feature could encompass making a (edited) status for the current preset view when these filters are changed while remaining in the preset view.
However, this would be in a separate PR, I have no issue with the current implementation thank you for clarifying your thought process.

nknguyenhc: I have found out an easy way to let the filters service track the labels fields as well. Because `sanitiseLabels` is only called whenever the labels are polled, I make sure that the update of filters from this method does not change the preset view value.

src/app/core/services/filters.service.ts
```diff
@@ -34,13 +35,47 @@ export const DEFAULT_FILTER: Filter = {
...
+    }),
+    custom: () => this.filter$.value
+  };
+
+  // List of keys in the new filter change that causes current filter to not qualify to be a preset view.
+  readonly presetChangingKeys = new Set<string>(['status', 'type', 'milestones']);
+
   public filter$ = new BehaviorSubject<Filter>(DEFAULT_FILTER);
+  // Either 'currentlyActive', 'contributions', or 'custom'.
+  public presetView$ = new BehaviorSubject<String>('custom');
```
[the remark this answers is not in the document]
nknguyenhc: Yes, fixed!

src/app/core/services/filters.service.ts
nknguyenhc: Fixed as well!

Arif-Khalid (approved):
I like the changes made and the current behaviour.
Great PR, LGTM!

NereusWB922 (dismissed):
LGTM!

NereusWB922 (approved):

PR #324, opened 2024-03-29
Reset GroupingContextService only if "keep filter" is selected.

NereusWB922:
### Summary:

Fix the bug where `GroupingContextService` is reset when the keep filter option is selected.

#### Type of change:

- 🐛 Bug Fix

### Changes Made:

- Add if block to check if keep filter option is selected before resetting

### Proposed Commit Message:

```
Reset GroupingContextService only if "keep filter" is selected.

GroupingContextService is being reset when the "keep filter" 
option is selected.

Let's check if the "keep filter" option is selected before
resetting the service.
```


Arif-Khalid (approved):
LGTM!

PR #362, opened 2024-04-10
Make ItemsPerPage common for all card views

### LINKED ISSUE #361: Make ItemsPerPage common to all card views
Currently, each card view has its own dropdown to select items per page. However, it is counterintuitive to have one column showing a different number of items on the page, as compared to another.

Move the `Items per page` dropdown into the filter bar so that it is commonly controlled for all card views.

**Describe the alternative you've considered**
To have the `Items per page` as a separate dropdown elsewhere on the page (for instance, on the bottom right), and implemented separately from the rest of the filters. However, to maintain uniformity of implementation, it is a part of the filter-bar

MadLamprey:
### Summary:

Fixes #361 

#### Type of change:

- ✨ Enhancement

### Changes Made:

- Add a filter for `Items per page`
- Create a subscription in `card-view.component` to the value of `itemsPerPage` for the `pageSize` property of the pagination

### Screenshots:

Screenshot 2024-04-11 at 12 04 35 AM

### Proposed Commit Message:

```
Make ItemsPerPage common for all card views

Currently, the items per page can be tweaked for each card view individually.
This is not intuitive, and it should be controlled commonly for all columns.

Let's make the items per page common for all card views.
```


src/app/shared/filter-bar/filter-bar.component.css
```diff
@@ -2,7 +2,7 @@
   margin: 8px;
   font-size: 14px;
   max-width: 20%;
-  width: 17%; /* depends on number of filters*/
+  width: 13%; /* depends on number of filters*/
```
nknguyenhc, line 5: Currently the dropdown filters are already quite cramp. Perhaps decrease the width of the search bar instead?

nknguyenhc (approved):
LGTM!

PR #378, opened 2024-06-24
Enhance error message on incorrect organisation/repo name

### LINKED ISSUE #376: Incorrect error message
From #367 

image

The above error message always appear whenever the repo is invalid. Sometimes it is due to org name, not repo name. The error message can be improved.

kokerinks:
### Summary:

Fixes #376 

#### Type of change:

- ✨ Enhancement

### Changes Made:

- Added a new error message `repoOwnerNotPresentMessage`, which will show now when the repo owner (i.e. organisation or username) in `OWNER_NAME/REPO` does not exist 
  - originally showed the message `repositoryNotPresentMessage` instead

### Screenshots:

In repo selection page:
image

After submitting:
image

Works for other repo-switching functions as well:
image
image

### Proposed Commit Message:

```
Enhance error message on incorrect organisation/repo name

The original error message for incorrect repo name does not specify 
whether the mistake is in the org name or repo name.

Hence, let's add an additional error message to show if the 
organisation name could not be found. This will give users an easier 
time finding the error and rectifying it when keying in a repo.
```

<details><summary>
<h3>Checklist:</h3>
</summary>

- [X] I have created tests for any new code files created in this PR or provided a link to a issue/PR that addresses this.
- [X] I have updated the project's documentation as necessary.

</details>

### DIFF
--- a/src/app/core/services/error-message.service.ts
+++ b/src/app/core/services/error-message.service.ts
@@ -8,12 +8,12 @@ import { Injectable } from '@angular/core';
  * Contains all error message prompts to user.
  */
 export class ErrorMessageService {
-  public static organisationNotPresentMessage() {
-    return 'Invalid organisation name. Please provide a repository URL or repo name (Org/Repo Name) with a valid organisation name.';
+  public static repoPrefixNotPresentMessage() {
+    return 'Invalid repo prefix. Please provide a repository URL or repo name with a valid prefix (org/user name), using the format <ORG/USER_NAME>/<REPO_NAME>';
   }
 
   public static repositoryNotPresentMessage() {
-    return 'Invalid repository name. Please provide Github repository URL or the repository name in the format Org/Repository Name.';
+    return 'Invalid repository name. Please provide Github repository URL or the repository name in the format <ORG/USER_NAME>/<REPO_NAME>.';
   }
 
   public static invalidUrlMessage() {
--- a/src/app/core/services/error-message.service.ts
+++ b/src/app/core/services/error-message.service.ts
@@ -8,8 +8,8 @@ import { Injectable } from '@angular/core';
  * Contains all error message prompts to user.
  */
 export class ErrorMessageService {
-  public static repoPrefixNotPresentMessage() {
-    return 'Invalid repo prefix. Provide a repository URL or repo name with a valid prefix. (e.g. <ORG/USER>/<REPO>)';
+  public static repoOwnerNotPresentMessage() {
+    return 'Invalid repo owner. Provide a repository URL or repo name (<ORG/USER>/<REPO>) with a valid organisation/user name';
   }
 
   public static repositoryNotPresentMessage() {
--- a/src/app/core/services/error-message.service.ts
+++ b/src/app/core/services/error-message.service.ts
@@ -8,6 +8,10 @@ import { Injectable } from '@angular/core';
  * Contains all error message prompts to user.
  */
 export class ErrorMessageService {
+  public static organisationNotPresentMessage() {
+    return 'Invalid organisation name. Please provide a repository URL or repo name (Org/Repo Name) with a valid organisation name.';
+  }
+
   public static repositoryNotPresentMessage() {
     return 'Invalid repository name. Please provide Github repository URL or the repository name in the format Org/Repository Name.';
   }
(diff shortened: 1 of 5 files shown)


nknguyenhc:
@kokerinks If your PR is ready for review, do mark it as ready so that we can start the review process.

kokerinks:
@nknguyenhc marked as ready for review 👍

nknguyenhc (changes requested):
Nice job on following our coding convention!

image

`nknguyenhc` is a valid Github profile. You may want to check for whether the user exists as well, and change the error messages from "invalid organisation" to "invalid organisation/user"

kokerinks:
@nknguyenhc I have made the required changes, feel free to review again

src/app/issues-viewer/issues-viewer.component.ts
```diff
@@ -123,7 +123,10 @@ export class IssuesViewerComponent implements OnInit, AfterViewInit, OnDestroy {
       return of(false);
     }
 
-    return this.githubService.isRepositoryPresent(currentRepo.owner, currentRepo.name);
+    return (
+      this.githubService.isOrganisationPresent(currentRepo.owner) &&
+      this.githubService.isRepositoryPresent(currentRepo.owner, currentRepo.name)
```
nknguyenhc, line 128: This function is no longer used, but for the sake of correctness, you may want to include the check on user as well.

nknguyenhc (changes requested):
image

When a user is viewing a repo and changes to an invalid repo,

1. Expected behaviour: Error message pops up and user can continue viewing the previous valid repo. This is also the behaviour before this change.
2. Actual behaviour: Error message pops up, but the user cannot continue viewing the previous valid repo.

src/app/core/services/error-message.service.ts
```diff
@@ -8,8 +8,12 @@ import { Injectable } from '@angular/core';
  * Contains all error message prompts to user.
  */
 export class ErrorMessageService {
+  public static repoPrefixNotPresentMessage() {
+    return 'Invalid repo prefix. Provide a repository URL or repo name with a valid prefix. (e.g. <ORG/USER>/<REPO>)';
```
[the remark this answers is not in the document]
kokerinks: Hmm.. I thought it might be too clunky to name it `organisationOrUserNotPresentMessage`, thats why i tried using "repo prefix" instead, but I understand that the term may not be intuitive enough

Do you have any suggestions as to what we can name it? I was thinking maybe `repoOwnerNotPresentMessage`?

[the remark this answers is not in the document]
nknguyenhc: `repoOwnerNotPresentMessage` sounds good.

kokerinks:
Have addressed issues again in latest commits, let me know if there's any other changes to be made!

src/app/core/services/view.service.ts
```diff
@@ -114,18 +114,41 @@ export class ViewService {
...
+   * @throws Error if the repository is not valid
+   */
+  async changeRepositoryIfValid(repo: Repo) {
+    try {
+      this.isChangingRepo.next(true);
+      await this.verifyOwnerAndRepo(repo.owner, repo.name);
+    } catch (error) {
+      this.isChangingRepo.next(false);
+      throw error;
+    }
```
nknguyenhc, line 151: Would not recommend try-catching as part of the control flow, i.e. `this.isChangingRepo.next(false)` should not be in the `catch` block. Perhaps you can do this in `verifyOwnerAndRepo` before throwing the error in the same method, and remove the try-catch block here?

kokerinks: `verifyOwnerAndRepo` is used in another portion of the code though, which does not require the `this.isChangingRepo.next(false)` statement. 

Perhaps I can make `verifyOwnerAndRepo` return a constant string or enum? then i can throw the relevant errors from here instead. can do the same for the other usages of `verifyOwnerAndRepo` as well

nknguyenhc: I believe it is fine to have that line of code in the method `initializeCurrentRepository`? Do try and let me know if it causes any unintended effect.

kokerinks: hmm yeah should be okay, have commited and pushed the change

nknguyenhc (approved):
LGTM!

Issue #342, opened 2024-04-01
Add filter for assignee


NereusWB922:
**Is your feature request related to a problem? Please describe.**
Currently, we can group issues and PRs by different criteria, not only by assignee. We should provide a filter for the assignee so users can filter the issue based on the assignee in different groupings.

kokerinks:
hi can i work on this issue? will see if there's an intuitive way to add the filter for assignees as stated in the issue

NereusWB922:
Sure, please go ahead and lmk if u have any question.

kokerinks:
@NereusWB922 I have created the PR above as a prepatory PR to the enhancement.

If this is PR okay I intend to use `this.assignees` in the new `AssigneeService` to create the filter options, and will attempt to follow the format which the milestone filter was implemented.

NereusWB922:
@kokerinks, Thanks for the pr. I will only be available to review prs during the weekend. If you are confident enough, you can proceed to implement the remaining logic. Just remember to checkout a new branch from ur  feature branch in case any changes are required for #379.

kokerinks:
@NereusWB922 I have managed to get a preliminary assignee filter working, but I'd like to clarify on the expected behaviour here

Would you like the filter to filter based on the list of assignees in each issue/PR, or is it instead based on creator of each issue/PR?

My current implementation filters based on the former, and the behaviour doesnt seem to be what is expected:
<img width="1710" alt="image" src="https://github.com/CATcher-org/WATcher/assets/53986507/a77f2b34-3158-4510-a11e-126f79997066">
in this case, the list does show only unassigned issues/PRs, but there are still entries on some team members as they are unassigned issues who still have a creator.

NereusWB922:
Hi @kokerinks,

Currently, in WATcher, we mainly focus on visualizing the feature/enhancement contributions of students to the project. This filter is primarily for tutors or other team members to easily identify the issues fixed by the students and the PRs created by the students.

I agree that the term "assignee" is a bit confusing as it refers to different things for GitHub Issues and PRs. Another alternative I considered when creating this issue was "Contributor," but that can also be misleading. If you have any recommendations, please suggest them.

To summarize again, the filtering logic should be:

- Filter by assignee for Issues
- Filter by author for PRs

PS (not related to this issue):
[...]

PR #379, opened 2024-07-01
Refactor GithubUser model to no longer encapsulate assignee role

kokerinks:
### Summary:

This PR is in preparation for the enhancement proposed in the issue #342 

Before this PR, the `GithubUser` model was responsible for storing information for both the authentication as well as details of assignees in each issue/PR. This PR turns them into two separate classes, keeping `GitHubUser` for authentication and moving the details of assignees to a new class `Assignee`.

This is done to ensure separation of roles between models. This will also improve code quality for future PRs related to the role of `Assignees`.

#### Type of change:

- 🎨 Code Refactoring

### Changes Made:

- Created a new model `Assignee`
  - Similar in structure to the existing `Milestone` model
- Replaced all instances where `GithubUser` is used as an assignee with `Assignee`
  - Significant changes were in `assignee-grouping-strategy`, where GithubUser is used heavily as an assignee

### Proposed Commit Message:

```
Refactor GithubUser model to no longer encapsulate assignee role

Currently, GithubUser is responsible for two roles. One, it holds data
for the current user during and after Github authentication. Second, it
stores data for assignees of issues and PRs in the Github repository.

This commit refactors GithubUser to no longer encapsulate the assignee
role. Instead, it introduces a new class, Assignee, to encapsulate
assignee data. 

This change is important to support better cohesion of the components,
and facilitates easier maintenance and further development in both 
functions.
```

<details><summary>
<h3>Checklist:</h3>
</summary>

- [X] I have created tests for any new code files created in this PR or provided a link to a issue/PR that addresses this.
- [X] I have updated the project's documentation as necessary.

</details>

### DIFF
--- a/src/app/core/models/assignee.model.ts
+++ b/src/app/core/models/assignee.model.ts
@@ -4,7 +4,7 @@ import { Group } from './github/group.interface';
  * Represents an assignee and its attributes fetched from Github.
  */
 export class Assignee implements Group {
-  static WithoutAssignee: Assignee = new Assignee({ login: 'Issue/PR without an assignee', avatar_url: null });
+  static WithoutAssignee: Assignee = new Assignee({ login: 'Issues/PRs without an assignee', avatar_url: null });
   login: string;
   avatarUrl: string;
 
--- a/src/app/core/models/assignee.model.ts
+++ b/src/app/core/models/assignee.model.ts
@@ -1,0 +1,22 @@
+import { Group } from './github/group.interface';
+
+/**
+ * Represents an assignee and its attributes fetched from Github.
+ */
+export class Assignee implements Group {
+  static WithoutAssignee: Assignee = new Assignee({ login: 'Issue/PR without an assignee', avatar_url: null });
+  login: string;
+  avatarUrl: string;
+
+  constructor(assignee: { login: string; avatar_url: string }) {
+    this.login = assignee.login;
+    this.avatarUrl = assignee.avatar_url;
+  }
+
+  public equals(other: any) {
+    if (!(other instanceof Assignee)) {
+      return false;
+    }
+    return this.login === other.login;
+  }
+}
--- a/src/app/core/services/grouping/assignee-grouping-strategy.service.ts
+++ b/src/app/core/services/grouping/assignee-grouping-strategy.service.ts
@@ -5,6 +5,8 @@ import { GithubUser } from '../../models/github-user.model';
 import { Issue } from '../../models/issue.model';
 import { GithubService } from '../github.service';
 import { GroupingStrategy } from './grouping-strategy.interface';
+import { AssigneeService } from '../assignee.service';
+import { Assignee } from '../../models/assignee.model';
 
 /**
  * A GroupingStrategy that groups issues/prs based on their assignees.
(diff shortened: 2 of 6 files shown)


src/app/core/services/grouping/assignee-grouping-strategy.service.ts
```diff
@@ -13,30 +15,29 @@ import { GroupingStrategy } from './grouping-strategy.interface';
...
-    if (key === GithubUser.NO_ASSIGNEE) {
-      return this.getUnassignedData(issues);
-    }
-
-    return this.getDataAssignedToUser(issues, key);
+  getDataForGroup(issues: Issue[], key: Assignee): Issue[] {
+    return issues.filter(
+      (issue) => issue.assignees.includes(key.login) || (key === Assignee.WithoutAssignee && issue.assignees.length === 0)
+    );
   }
```
NereusWB922, line 29: In WATcher, the issue model can represent either a GitHub Issue or a GitHub Pull Request. Therefore, in getDataAssignedToUser, we use different filtering logic: for GitHub Issues, we retrieve those assigned to the target; for GitHub Pull Requests, we retrieve those created by the target.

src/app/core/models/assignee.model.ts
```diff
@@ -0,0 +1,22 @@
...
+    this.avatarUrl = assignee.avatar_url;
+  }
+
+  public equals(other: any) {
+    if (!(other instanceof Assignee)) {
+      return false;
+    }
+    return this.login === other.login;
+  }
+}
```
NereusWB922, line 22: I don't see the need for a new class to represent an assignee since an assignee can only be a GitHub User. Unless the assignee has different behaviors or fields from a GitHub user, creating a new class seems unnecessary. Additionally, you created a new service for the assignee, which just parses data from the GitHub user. 

If the concern is about not passing the entire GitHub user object to methods that only require the "login" and "avatarUrl" fields, a simpler solution would be to create an interface called Assignee with these fields and have the GitHub user implement this interface. This way, methods requiring an assignee can use the Assignee type, ensuring only the assignee-specific fields are accessed.

I might be wrong about this, so please feel free to discuss further if you have different considerations.

NereusWB922 (changes requested):
Hi @kokerinks, l left some comments, pls lmk if you have any questions.

src/app/core/models/assignee.model.ts
kokerinks: @NereusWB922 okay, after working deeper on the filter PR I think i have a better understanding of the role of each of the models and services.. will try to incorporate the new filter without using a new assignee model

that said, having a new service for assignee is fine right? I'll just rewrite it so that it uses githubUsers instead of the new assignee model class

NereusWB922: Hi @kokerinks, then in this case, you can just use the method getUsersAssignable from github service right? New service seems unnecessary also.

kokerinks: hmm i think it would make the code less easy to understand though. i intentionally designed `assignee.service.ts` to be similar in structure to `milestone.service.ts`, since they both have a similar purpose of handling logic for their respective filters.`assignee.service.ts` currently also "caches" the assignees taken from `getUsersAssignable`, which will help to reduce calls to github APIs.

There are some options i see here:
1. Use `getUsersAssignable` to retrieve each time and without any caching 
  - will cause structure of `assignee` filter implementation to differ from milestone
  - will result in more github api calls
2. Implement caching in `github.service.ts` 
  - code structure would not be consistent with `milestone.service.ts`, where the milestones are "cached" at
3. Implement caching in a new `assignee.service.ts`
  - will result in a new file, but will align better with the current milestone filter implementation in terms of code structure
  
Personally, I think having the new file is necessary.. i understand the potential concern of having too many service files, but maybe we can reorganise the services into new folders to declutter the `services` folder.

bit of a read, let me know what you think!

NereusWB922: Hi @kokerinks, thanks for the detailed explanation. I can see where you are coming from and agree that having the filtering logic in the GitHub User model would make it too complicated. After reviewing our codebase, I think our current filter service is too complicated and definitely needs some refactoring. For now, you can stick with your implementation. I will create a separate GitHub issue regarding the refactoring of the filter service.

NereusWB922: Just let me know when this pr is ready for review again. Thanks

kokerinks:
Reimplementing it in https://github.com/CATcher-org/WATcher/pull/381 alongside the assignee filter implementation

PR #382, opened 2024-07-14
Responsive scrollable filter bar

### LINKED ISSUE #377: Responsiveness of filters
When the screen width is reduced, the filter bars are cut off and not a responsive design

A togglable hamburger when the width reaches a certain minimum that houses filters

Making the filters on top of one another is also possible but might create a really long page

**Additional context**
Before reducing screen width
image

After reducing screen width
image

JiaXinEu:
### Summary:

Fixes #377 

#### Type of change:

- ✨ Enhancement

### Changes Made:

- Add `BreakpointObserver` to observe and respond to screen size changes
- Add private method to update styling of `filter-bar`
- Put `dropdown-filters` in scrollable container

### Screenshots:

image

image

### Proposed Commit Message:

```
Modify responsiveness of filters to screen size change

The filter bar is cut off when screen width is reduced. We should 
make the design responsive.

Let's make filter bar scrollable when it cannot fit on screen.
```

<details><summary>
<h3>Checklist:</h3>
</summary>

</details>

### DIFF
--- a/src/app/shared/filter-bar/filter-bar.component.css
+++ b/src/app/shared/filter-bar/filter-bar.component.css
@@ -16,11 +16,6 @@ ul {
   max-height: 80px;
 }
 
-.search-bar {
-  width: 100%;
-  flex-grow: 1;
-}
-
 .menu {
   padding: 0 10px;
   border-left: 1px solid #eee;
@@ -38,7 +33,6 @@ ul {
   display: flex;
   align-items: center;
   flex: 1;
-  flex-grow: 3;
   margin-right: 100px;
 
   @media (max-width: 1400px) {
@@ -57,9 +51,5 @@ ul {
         margin: 8px;
       }
     }
-  }
-
-  .filter-item {
-    display: block;
   }
 }
--- a/src/app/shared/filter-bar/filter-bar.component.css
+++ b/src/app/shared/filter-bar/filter-bar.component.css
@@ -12,19 +12,13 @@ ul {
   align-items: center;
   justify-content: space-between;
   list-style-type: none;
+  position: relative;
+  max-height: 80px;
 }
 
 .search-bar {
   width: 100%;
-}
-
-.filters {
-  width: 57%;
-  display: flex;
-}
-
-.label-filter-grid-tile {
-  width: 100%;
+  flex-grow: 1;
 }
 
 .menu {
@@ -33,35 +27,39 @@ ul {
   display: none;
   cursor: pointer;
 
-  @media (max-width: 1148px) {
+  @media (max-width: 1400px) {
     display: block;
     justify-items: right;
   }
 }
 
 .filters {
+  width: 57%;
   display: flex;
   align-items: center;
+  flex: 1;
+  flex-grow: 3;
+  margin-right: 100px;
 
-  @media (max-width: 1148px) {
+  @media (max-width: 1400px) {
     display: none;
 
     &.is-open {
       display: block;
       position: absolute;
       width: 100%;
-      top: calc(20% + 8px); /* 8px margin from ul */
+      top: 100%;
       z-index: 10;
-      background-color: #fffcfc;
+      background-color: #ffffff;
 
       .filter-item {
         display: inline-block;
+        margin: 8px;
       }
     }
   }
 
   .filter-item {
-    padding: 0 10px;
     display: block;
   }
 }
--- a/src/app/shared/filter-bar/filter-bar.component.css
+++ b/src/app/shared/filter-bar/filter-bar.component.css
@@ -19,7 +19,7 @@
 
 .label-filter-grid-tile {
   width: 100%;
-  justify-content: flex-end;
+  justify-content: end;
 }
 
 .submit-button {
@@ -31,7 +31,7 @@
   overflow-x: auto;
   display: flex;
   justify-content: center;
-  scrollbar-width: 2px;
+  scrollbar-width: 5px;
 }
 
 @media (max-width: 960px) {
--- a/src/app/shared/filter-bar/filter-bar.component.css
+++ b/src/app/shared/filter-bar/filter-bar.component.css
@@ -1,41 +1,67 @@
-.dropdown-filters .mat-form-field {
+.filters .mat-form-field {
   margin: 8px;
   font-size: 14px;
   max-width: 20%;
   width: 14%; /* depends on number of filters*/
-  flex-shrink: 0;
-  min-width: 110px;
+}
+
+ul {
+  margin: 8px;
+  padding: 0;
+  display: flex;
+  align-items: center;
+  justify-content: space-between;
+  list-style-type: none;
 }
 
 .search-bar {
-  width: 80%;
+  width: 100%;
 }
 
-.dropdown-filters {
-  width: 100%;
+.filters {
+  width: 57%;
   display: flex;
-  flex-wrap: nowrap;
 }
 
 .label-filter-grid-tile {
   width: 100%;
-  justify-content: end;
 }
 
-.submit-button {
-  margin-left: 8px;
+.menu {
+  padding: 0 10px;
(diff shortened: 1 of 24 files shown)


src/app/shared/filter-bar/filter-bar.component.ts
```diff
@@ -55,11 +70,22 @@ export class FilterBarComponent implements OnInit, OnDestroy {
...
+
+    this.breakpointSubscription = this.breakpointObserver.observe([Breakpoints.Small, Breakpoints.XSmall]).subscribe((result) => {
+      this.isSmallScreen = result.matches;
+      this.updateSpan();
+    });
+
+    this.breakpointSubscription = this.breakpointObserver.observe([Breakpoints.Medium]).subscribe((result) => {
+      this.isMediumScreen = result.matches;
+      this.updateSpan();
+    });
```
Arif-Khalid, line 82: While this works, would it be better to take advantage of the auto scaling capabilities of css? 
We can avoid these breakpoints and corresponding values since the parent colspan of 7 scales according to screen size already, such that the children widths are always relative to the parent without changing their individual colspan or the parents. 
In order to preserve the full child item being shown when the screen shrinks, consider instead using min-width/max-width css properties with media queries to shift responsiveness logic to css where possible.

src/app/shared/filter-bar/filter-bar.component.css
```diff
@@ -11,12 +13,29 @@
...
 .submit-button {
   margin-left: 8px;
 }
+
+.dropdown-filters-container {
+  margin-right: 8px;
+  overflow-x: auto;
+  display: flex;
+  justify-content: center;
+  scrollbar-width: 5px;
```
Arif-Khalid, line 34: Should this be using the firefox specific styling? 
It would be better to use webkit:scrollbars pseudo-class selector in addition to this to properly style scrollbars in all browsers.
See https://stackoverflow.com/questions/15420314/how-to-make-custom-scrollbars-show-in-all-browsers

src/app/shared/filter-bar/filter-bar.component.css
```diff
@@ -11,12 +13,29 @@
...
 
 .submit-button {
   margin-left: 8px;
 }
+
+.dropdown-filters-container {
+  margin-right: 8px;
+  overflow-x: auto;
+  display: flex;
+  justify-content: center;
```
Arif-Khalid, line 33: Is there a reason for justify-content and changing it when the screen is smaller? There seems to be no difference if I remove justify content in both classes in testing.

Arif-Khalid (changes requested):
I see the vision and am open to the possibility of using a scrollable filters container like you have implemented. 
However, I feel that the scrollbar is not very intuitive to the user, i.e., not what you would find on most other sites. It will also be quite hard to use on a small screen like a mobile screen compared to a hamburger or dropdown menu, which is the very audience this change is supposed to cater to.
I would prefer a different approach but you are free to continue working on this approach and try to apply some cosideration for mobile users who don't have that much space to scroll horizontally for example left and right scroll buttons at the ends of the scrollbar.

JiaXinEu:
> I see the vision and am open to the possibility of using a scrollable filters container like you have implemented. However, I feel that the scrollbar is not very intuitive to the user, i.e., not what you would find on most other sites. It will also be quite hard to use on a small screen like a mobile screen compared to a hamburger or dropdown menu, which is the very audience this change is supposed to cater to. I would prefer a different approach but you are free to continue working on this approach and try to apply some cosideration for mobile users who don't have that much space to scroll horizontally for example left and right scroll buttons at the ends of the scrollbar.

Thank you for a better insight of the issue. The current PR uses a togglable menu approach

src/app/shared/filter-bar/filter-bar.component.css
```diff
@@ -1,22 +1,67 @@
...
+  align-items: center;
+
+  @media (max-width: 1148px) {
+    display: none;
+
+    &.is-open {
+      display: block;
+      position: absolute;
+      width: 100%;
+      top: calc(20% + 8px); /* 8px margin from ul */
```
Arif-Khalid, line 53: Would it be better to use a calculation relative to the app-filter-bar? By applying a position: relative to the app filter bar you can position the child element of class filters relative to that parent when the child has position: absolute. Therefore you could then do something like top: 100%; instead on this element and it will be positioned nicely under the filter bar.
image

Arif-Khalid: Could you share your reasoning behind the 8px margin from ul as commented? I find it uneccessary and it would be better to associate the expansion more closely with the app-filter-bar.

src/app/shared/filter-bar/filter-bar.component.css
```diff
@@ -1,22 +1,67 @@
...
 
-.search-bar {
-  width: 80%;
+ul {
+  margin: 8px;
+  padding: 0;
+  display: flex;
+  align-items: center;
+  justify-content: space-between;
+  list-style-type: none;
```
Arif-Khalid, line 14: Good substitution for the mat-grid components. Could you enforce a height of 80px so that no vertical scrollbar appears ? This was was enforced by the previous rowHeight property.

src/app/shared/filter-bar/filter-bar.component.css
```diff
@@ -1,22 +1,67 @@
...
 }
 
-.dropdown-filters {
+.search-bar {
```
Arif-Khalid, line 17: Is this class necessary?

src/app/shared/filter-bar/filter-bar.component.css
```diff
@@ -1,22 +1,67 @@
...
 
-.submit-button {
-  margin-left: 8px;
+.menu {
+  padding: 0 10px;
+  border-left: 1px solid #eee;
+  display: none;
+  cursor: pointer;
+
+  @media (max-width: 1148px) {
```
Arif-Khalid, line 36: At about 1200-1300 px range there is overlapping of components
image
Would it be better to use a more standard, larger breakpoint of 1400px?
Refer to https://getbootstrap.com/docs/5.0/layout/breakpoints/ for some standard breakpoints.

src/app/shared/filter-bar/filter-bar.component.css
```diff
@@ -1,22 +1,67 @@
...
+    display: block;
+    justify-items: right;
+  }
+}
+
+.filters {
+  display: flex;
+  align-items: center;
+
+  @media (max-width: 1148px) {
```
Arif-Khalid, line 46: Same comment about breakpoints

src/app/shared/filter-bar/filter-bar.component.css
```diff
@@ -1,22 +1,67 @@
...
+
+      .filter-item {
+        display: inline-block;
+      }
+    }
+  }
+
+  .filter-item {
+    padding: 0 10px;
+    display: block;
```
Arif-Khalid, line 65: Is there a need for display: block and display: inline-block? Children of flexbox are automatically given display: inline-block

Arif-Khalid (changes requested):
Great implementation of a hamburger dropdown and I like the substitution of mat grid you made.
However, there is some redundancy and minor errors in regards to CSS to fix.
Additionally, would it be better to put the app-label-filters into the hamburger dropdown as well since on small screens at first glance it now seems to suggest it denotes something about all filters.

JiaXinEu:
> Additionally, would it be better to put the app-label-filters into the hamburger dropdown as well since on small screens at first glance it now seems to suggest it denotes something about all filters.

Thank you for the feedback. I have moved label-filter into the hamburger dropdown as well, and done the minor fixes on css as suggested.

Arif-Khalid (approved):
Your changes make the app a lot more usable on smaller devices. Thanks so much for your contributions!
LGTM!

Issue #385, opened 2024-08-03
Incorrect filtering of issues with multiple assignees


kokerinks:
**Describe the bug**
Bug from PR #381 , new assignee filter breaks when issues have more than one assignee

**To Reproduce**

1. Go to repository with an issue containing multiple assignees
1. Ensure the issues/PRs are currently grouped by assignees
1. Under assignee filter, select only one of the assignees for that issue
1. Notice that additional users and issues are shown, even if that user is not included in the assignee filter
![20240803-171558](https://github.com/user-attachments/assets/29926285-47bb-4279-9f3f-0fa062ec08f1)

**Expected behavior**
Only columns of users included in the assignee filter should be displayed.

**Desktop (please complete the following information):**

- OS: Windows 11
- Browser: MS Edge
- Version: 127.0.2651.74

kokerinks:
Sorry have been very busy, thank you @IzN432 @Eclipse-Dominator @Arif-Khalid for the help!

PR #389, opened 2025-01-12
Fix incorrect assignee filtering  

kokerinks:
### Summary:
Fixes bug #385 

#### Type of change:

- 🐛 Bug Fix

### Changes Made:
Assignee filter now correctly removes Assignee columns when it is not selected in the filter:
image

### Proposed Commit Message:

```
Fix incorrect filtering of issues with multiple assignees

Previously, the application will sometimes display additional assignees
which were not selected in the assignee filter.

Let's fix this bug and ensure that the application only shows for 
assignees selected in the assignee filter.
```

<details><summary>
<h3>Checklist:</h3>
</summary>

</details>

### DIFF
--- a/src/app/core/services/filters.service.ts
+++ b/src/app/core/services/filters.service.ts
@@ -359,4 +359,8 @@ export class FiltersService {
     // TODO Filter out assignees that have not contributed in currently active milestones
     return [...this.assigneeService.assignees, GithubUser.NO_ASSIGNEE];
   }
+
+  getAssignees(): GithubUser[] {
+    return this.assigneeService.assignees.filter((assignee) => this.filter$.value.assignees.includes(assignee.login));
+  }
 }
--- a/src/app/core/services/grouping/assignee-grouping-strategy.service.ts
+++ b/src/app/core/services/grouping/assignee-grouping-strategy.service.ts
@@ -3,6 +3,7 @@ import { Observable } from 'rxjs';
 import { map } from 'rxjs/operators';
 import { GithubUser } from '../../models/github-user.model';
 import { Issue } from '../../models/issue.model';
+import { FiltersService } from '../filters.service';
 import { GithubService } from '../github.service';
 import { GroupingStrategy } from './grouping-strategy.interface';
 
@@ -13,7 +14,7 @@ import { GroupingStrategy } from './grouping-strategy.interface';
   providedIn: 'root'
 })
 export class AssigneeGroupingStrategy implements GroupingStrategy {
-  constructor(private githubService: GithubService) {}
+  constructor(private githubService: GithubService, private filtersService: FiltersService) {}
 
   /**
    * Retrieves data for a specific assignee.
@@ -45,7 +46,13 @@ export class AssigneeGroupingStrategy implements GroupingStrategy {
    * hidden group list if empty.
    */
   isInHiddenList(group: GithubUser): boolean {
-    return group !== GithubUser.NO_ASSIGNEE;
+    return (
+      group !== GithubUser.NO_ASSIGNEE && // group is not "No Assignee"
+      !this.filtersService
+        .getAssignees()
+        .map((assignee) => assignee.login)
+        .includes(group.login)
+    ); // group is not selected
   }
 
   private getDataAssignedToUser(issues: Issue[], user: GithubUser): Issue[] {
--- a/src/app/issues-viewer/issues-viewer.component.html
+++ b/src/app/issues-viewer/issues-viewer.component.html
@@ -11,7 +11,9 @@
         *ngFor="let group of this.groupService.groups"
         class="issue-table"
         #card
-        [ngStyle]="{ display: card.isLoading || card.issueLength > 0 ? 'initial' : 'none' }"
+        [ngStyle]="{
+          display: card.isLoading || (card.issueLength > 0 && !groupService.hiddenGroups.includes(card.group)) ? 'initial' : 'none'
+        }"
         [group]="group"
         [headers]="this.displayedColumns"
         (issueLengthChange)="this.groupService.updateHiddenGroups($event, group)"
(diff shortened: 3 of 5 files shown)


kokerinks:
@Arif-Khalid @Eclipse-Dominator please help to review whenever convenient, thanks!

src/app/core/services/grouping/assignee-grouping-strategy.service.ts
```diff
@@ -45,7 +46,13 @@ export class AssigneeGroupingStrategy implements GroupingStrategy {
...
    */
   isInHiddenList(group: GithubUser): boolean {
-    return group !== GithubUser.NO_ASSIGNEE;
+    return (
+      group !== GithubUser.NO_ASSIGNEE && // group is not "No Assignee"
+      !this.filtersService
+        .getAssignees()
+        .map((assignee) => assignee.login)
+        .includes(group.login)
+    ); // group is not selected
```
Arif-Khalid, line 55: Should groups be forced into the hidden list? 
The functionality of the hidden list is to display information about which groups happen to be hidden because of the filters selected. 
The use of this function to specify unique behaviour for groups that shouldn't be shown in the hidden list even when they are empty, such as the NO_ASSIGNEE group.
Forcing groups into the hidden list goes against this single responsibility and erodes into the responsibility of the filters.

src/app/issues-viewer/issues-viewer.component.html
```diff
@@ -11,7 +11,9 @@
         *ngFor="let group of this.groupService.groups"
         class="issue-table"
         #card
-        [ngStyle]="{ display: card.isLoading || card.issueLength > 0 ? 'initial' : 'none' }"
+        [ngStyle]="{
+          display: card.isLoading || (card.issueLength > 0 && !groupService.hiddenGroups.includes(card.group)) ? 'initial' : 'none'
+        }"
```
Arif-Khalid, line 16: Should the group service be used as a condition to hide cards? 
Hiding based on the group service erodes into the function of filtering where group service is only meant to group, not filter.

src/app/core/services/filters.service.ts
```diff
@@ -359,4 +359,8 @@ export class FiltersService {
     // TODO Filter out assignees that have not contributed in currently active milestones
     return [...this.assigneeService.assignees, GithubUser.NO_ASSIGNEE];
   }
+
+  getAssignees(): GithubUser[] {
+    return this.assigneeService.assignees.filter((assignee) => this.filter$.value.assignees.includes(assignee.login));
+  }
```
Arif-Khalid, line 365: Should the filtersService contain a method to return a list of filtered assignees?
The filters service has the single responsibility of housing filters used throughout the application. Therefore the most updated filters should already be within the filter$ observable, handled by the component updating the filter. This method goes against this responsibility and its implementation even outside the filters service shouldn't happen since it implies an error in the use of the filters service.

Arif-Khalid: I understand your following of the already implemented methods ``getAssigneesForCurrentlyActive`` and ``getMilestonesForCurrentlyActive``. Those methods are exclusively used to populate the filters based on the preset view which is currently handled by the filters service. 
As an aside, it might make more sense to move the preset view and the corresponding functions out of filters service in the future.

Arif-Khalid (changes requested):
A good effort at eradicating the bug that works effectively. 
However, a number of responsibilities of the services are eroded by the implementation.
I would suggest a much simpler and structured fix by looking at ``~/src/app/shared/issue-tables/dropdownfilter.ts``. It already has reference to both services and has a consolidated section where filters are applied.
Please follow up with any questions or concerns you might have.

kokerinks:
@Arif-Khalid Thank you for your review. 

Before this PR, I did attempt a solution through the aforementioned `~/src/app/shared/issue-tables/dropdownfilter.ts`. However, I noticed it may not be possible due to the following reasons:
1. Unlike Milestones, each PR/issue can belong to more than one Assignee
2. Also note that the existing `applyDropdownFilter` implementation filters by **issue/PR, not by issue-card\*\***
3. Hence, what happens is:
    - Assume an issue X has two assignees A and B, and assignee A is selected on the filter while B is not
    - `applyDropdownFilter` filters by issue/PR, and accepts issue X as A is selected within the assignee filter
    - However, when displaying the issue-cards, the system detects that issue X has two assignees A and B, and displays issue-cards under the assignee columns A and B separately (even though it should only display an issue-card for A, and not B)

However, I do agree that this PR's solution is indeed a bit hard-coded and rather unstructured. I will try to look deeper into it, if you have any other suggestions on how to proceed as well please do let me know, or feel free to contact me on the slack channel. Thanks!

\*\* For clarity's sake, this is what i am referring to as an issue-card:

Eclipse-Dominator (changes requested):
The filtered items breaks when grouping by milestone.
image


This can be solved by including back the `issueLength == 0` check but will
in turn break grouping by user check. Therefore, this might not be a good solution to solve this bug.

---

grouping service updates their hidden and displayed groups from result of IssueDataTable filters. Since this issue is present only for the case of multiple assignees., It is probably best to approach from the filters.

Some initial ideas that come to mind are:
let IssueDataTable obtain a list of hidden assignees (from some service probably)

```typescript
   this.issueSubscription = merge(...displayDataChanges)
      .pipe(
        // maps each change in display value to new issue ordering or filtering
        map(() => {
          let data = <Issue[]>Object.values(this.issueService.issues$.getValue()).reverse();
          if (this.defaultFilter) {
            data = data.filter(this.defaultFilter);
          }
          /* if group.equals matches any value in hidden assignees
                  => immediately return [] as this datatable is guarenteed to be hidden
          */
          // ...
          // Filter by assignee of issue
          data = this.groupingContextService.getDataForGroup(data, this.group);

```
We can guarentee its correctness as group.equals will only return true when group is of type GithubUser type implying that it is grouping via assignees. Therefore, doing so will cause IssueDataTable to emit 0 length which means means the given group will be hidden by `updateHiddenGroups` from `group.service.ts`.

PR #417, opened 2025-02-25
Align remove all button to expectations of function

### LINKED ISSUE #398: Remove all button doesn't work
**Describe the bug**
The 'remove all' button under labels does not work. 

**To Reproduce**

1. Open WATcher
2. Click on the label filter:

Image

3. Click on the remove all button:

Image

4. Nothing happens

**Expected behavior**
All labels should be removed

- OS: Windows 11
- Browser: Chrome
- Version: 132.0.6834.84

HollaG:
### Summary:

Fixes #398 

#### Type of change:
- ✨ New Enhancement
- 🐛 Bug Fix
- 🧪 Tests Update

### Changes Made:
Previously, the "remove all" button was ambigious in its function. Let's move to refactor it such that it resets all modifications made to the filters instead.

### Screenshots:

https://github.com/user-attachments/assets/7be7ede1-fcf8-425c-9017-e1217818c357

### Proposed Commit Message:

```
Fix remove all button

The current "remove all" button might
lead to confusion in its function, as it doesn't
"remove" all changes made to the label selection.

Let's rename the button to "reset" to make it more
clear, and fix its function such that it resets all 
changes instead.
```

<details><summary>
<h3>Checklist:</h3>
</summary>

- [x] I have created tests for any new code files created in this PR or provided a link to a issue/PR that addresses this.
- [x] I have updated the project's documentation as necessary.

</details>

### DIFF
--- a/src/app/shared/filter-bar/label-filter-bar/label-filter-bar.component.ts
+++ b/src/app/shared/filter-bar/label-filter-bar/label-filter-bar.component.ts
@@ -117,13 +117,15 @@ export class LabelFilterBarComponent implements OnInit, AfterViewInit, OnDestroy
   updateSelection(): void {
     this.filtersService.updateFilters({
       labels: Array.from(this.selectedLabelNames),
-      deselectedLabels: this.deselectedLabelNames
+      deselectedLabels: this.deselectedLabelNames,
+      hiddenLabels: this.hiddenLabelNames
     });
   }
 
-  removeAllSelection(): void {
+  resetSelection(): void {
     this.selectedLabelNames = new Set<string>();
     this.deselectedLabelNames = new Set<string>();
+    this.hiddenLabelNames = new Set<string>();
     this.updateSelection();
   }
 }
--- a/src/app/shared/filter-bar/label-filter-bar/label-filter-bar.component.ts
+++ b/src/app/shared/filter-bar/label-filter-bar/label-filter-bar.component.ts
@@ -44,6 +44,7 @@ export class LabelFilterBarComponent implements OnInit, AfterViewInit, OnDestroy
         this.hiddenLabelNames = this.filtersService.filter$.value.hiddenLabels;
         this.loaded = true;
 
+        // set the default based on the initial round of filters
         this.isDefault = this.selectedLabelNames.size === 0 && this.deselectedLabelNames.size === 0 && this.hiddenLabelNames.size === 0;
       });
     });
@@ -59,8 +60,7 @@ export class LabelFilterBarComponent implements OnInit, AfterViewInit, OnDestroy
     }
 
     this.hiddenLabelNames.add(label);
-    this.isDefault = false;
-    this.filtersService.updateFilters({ hiddenLabels: this.hiddenLabelNames });
+    this.updateSelection();
   }
 
   /** Show labels that were hidden */
@@ -70,7 +70,6 @@ export class LabelFilterBarComponent implements OnInit, AfterViewInit, OnDestroy
     }
     this.hiddenLabelNames.delete(label);
     this.updateSelection();
-    this.isDefault = this.selectedLabelNames.size === 0 && this.deselectedLabelNames.size === 0 && this.hiddenLabelNames.size === 0;
   }
 
   /**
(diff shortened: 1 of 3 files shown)


IzN432 (approved):
LGTM! Did you decide not to disable the button when nothing is selected?

src/app/shared/filter-bar/label-filter-bar/label-filter-bar.component.ts
```diff
@@ -81,6 +87,7 @@ export class LabelFilterBarComponent implements OnInit, AfterViewInit, OnDestroy
     } else {
       this.selectedLabelNames.add(label.name);
     }
+    this.isDefault = this.selectedLabelNames.size === 0 && this.deselectedLabelNames.size === 0 && this.hiddenLabelNames.size === 0;
```
Eclipse-Dominator, line 90: Setting default appeared quite a lot of time. Might as well add a private functions that does this.
e.g.
```typescript
private updateIsDefault(): void {
    this.isDefault = this.selectedLabelNames.size === 0 && this.deselectedLabelNames.size === 0 && this.hiddenLabelNames.size === 0
}
```

Since we are only checking size of 3 arrays, this does not incur too much overhead when compared to just doing this.isDefault = true/false. We trade off a small bit of function call overhead for a more readable and maintainable code.

What do you think?

Eclipse-Dominator (changes requested):
Except for a small bit of refactoring, LGTM!

Eclipse-Dominator:
Small bit of update, I noticed `hide(string)` does not use `updateSelection` and I think they do roughly the same thing. 

Might be better if we standardize `hide` to use `updateSelection` and you only need to do the following once.
```
this.isDefault = this.selectedLabelNames.size === 0 && this.deselectedLabelNames.size === 0 && this.hiddenLabelNames.size === 0;
```

Which means you dont really have to add a new function as mentioned in the comment.

src/app/shared/filter-bar/label-filter-bar/label-filter-bar.component.ts
HollaG: I had actually considered that, one (alternative) thing which I tried was to have the 
`this.isDefault = this.selectedLabelNames.size === 0 && this.deselectedLabelNames.size === 0 && this.hiddenLabelNames.size === 0`

set in `updateSelection()` also,

but that broke all the test cases as for e.g. the test case for `hide` was only expecting `filterService.updateFilters` to be called with a change in `hiddenLabels`. I left it for now as I thought the test cases shouldn't really be changed.

In that way, we wouldn't need to have yet another function, but I wasn't sure whether I should be changing test cases.

HollaG:
> Small bit of update, I noticed `hide(string)` does not use `updateSelection` and I think they do roughly the same thing.
> 
> Might be better if we standardize `hide` to use `updateSelection` and you only need to do the following once in `updateSelection()` and you can delete the rest of `this.isDefault = ...`:
> 
> ```ts
> this.isDefault = this.selectedLabelNames.size === 0 && this.deselectedLabelNames.size === 0 && this.hiddenLabelNames.size === 0;
> ```
> 
> Which means you dont really have to add a new function as mentioned in the comment.

Oops, I didn't read this comment before replying to the review. As stated, this would require the test cases to be changed, is that ok?

Eclipse-Dominator:
> Oops, I didn't read this comment before replying to the review. As stated, this would require the test cases to be changed, is that ok?

I took a quick look at the unit test for show and hide. It occurs to me that the unit tests for show is not actually testing show properly. Hence, the breakage of show when `hide` was modified.

image
As you can see, unit test for `show` did not actually tests for certain behaviors of `show`. Hence, the unit tests right now is missing coverage for `show()`

As this appears to be another bug, I suggest pausing this PR and do a quick issue and PR to refactor filter-bar and improve unit test for show().

What do you think?

HollaG:
Looks good, do you suggest simply moving all instances of `this.filtersService.updateFilters` into `updateSelection?`

Eclipse-Dominator:
> This would be fine!

Eclipse-Dominator (approved):
Function and unit tests work as expected.

LGTM!

PR #420, opened 2025-03-01
Remove gap between bottom of window and horizontal sidebar using flexbox

### LINKED ISSUE #416: Remove gap between bottom of window and horizontal sidebar
**Describe the bug**
There is an obvious gap between the horizontal scroll bar(/ page navigation) and the bottom of the window. (refer to the screenshots)

I feel that ideally, the horizontal slidebar and page navigation should stick to the bottom of the window as it feels quite unnatural to see that gap since we dont have any elements below the current view. 

What do you guys think? @CATcher-org/2425s2 

**To Reproduce**
1. Go to WATcher in a browser with enough vertical height
2. Notice the presence of the gap mentioned above

**Expected behavior**
The gap shouldn't exist, and more space can be utilized.

**Screenshots**

Image

Image

wx-03:
### Summary:

Fixes #416 

#### Type of change:

- 🐛 Bug Fix

### Changes Made:

- Initially, each column was set to a height of 77vh in card-view component, which resulted in a weird gap between the bottom of the window and the horizontal sidebar on screens that have enough vertical height. 
- Setting the height to (100vh - filter bar height - header height) may not be a feasible solution as the heights will be hardcoded and need to be changed when the height of the filter bar / header changes.
- Thus I modified `app-root` to be a flexbox, which allows the content to take up the remaining vertical space of the viewport.  

### Screenshots:

Before:
image

After:
image

### Proposed Commit Message:

```
Remove gap between bottom of window and horizontal sidebar

The height of each column in `card-view` component is set to 77vh,
which results in a gap between the bottom of the window and the
horizontal sidebar. 

The content should fill up the remaining height of the window instead
of leaving a gap.

Let's use a flexbox to allow the content to dynamically change its
height to fill up the remaining space instead of having a fixed height.
```

<details><summary>
<h3>Checklist:</h3>
</summary>

</details>

### DIFF
--- a/src/app/app.component.css
+++ b/src/app/app.component.css
@@ -1,4 +1,5 @@
 .container {
   padding: 8px 16px;
   min-height: 87.9vh;
+  flex-grow: 1;
 }
--- a/src/app/app.component.html
+++ b/src/app/app.component.html
@@ -1,4 +1,4 @@
-<app-layout-header></app-layout-header>
+<app-layout-header style="flex: 0 0 auto"></app-layout-header>
 
 <div class="container">
   <router-outlet></router-outlet>
--- a/src/app/issues-viewer/card-view/card-view.component.css
+++ b/src/app/issues-viewer/card-view/card-view.component.css
@@ -1,6 +1,6 @@
 .card-column {
   margin: 8px;
-  height: 100vh;
+  height: calc(100vh - 192px);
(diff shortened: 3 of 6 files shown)


Eclipse-Dominator (approved):
Function works as expected. More space is now utilized!

LGTM!

